PR #70, opened 2017-08-18
remove the use of data helper from InstallSchema

BlackIkeEagle:
Problem:

When we do a clean magento2 install we get an exception caused by this
module.

  [Magento\Framework\Exception\LocalizedException]
  Invalid entity_type specified: customer

The reason for this is that de InstallData is executed after all
installations of schemas and updates of schemas.

Solution:

Remove the use of the helper in the InstallSchema, the helper will cause
the ObjectManager to instantiate several objects and that causes the
exception because the eav entity type for customer is not yet set.

The helper got removed from the InstallSchema, and due to that we use BP
now instead of the helpers getBasePath method. Additionally the try
catch to ignore the failure of creating the MailChimp folder in var is
removed. I think its better to fail hard there instead of ignoring the
fact the folder could not be created.

Signed-off-by: BlackEagle <ike.devolder@gmail.com>

### DIFF
--- a/Setup/InstallSchema.php
+++ b/Setup/InstallSchema.php
@@ -19,13 +19,6 @@ use Symfony\Component\Config\Definition\Exception\Exception;
 
 class InstallSchema implements InstallSchemaInterface
 {
-    private $_helper;
-
-    public function __construct(\Ebizmarts\MailChimp\Helper\Data $helper)
-    {
-        $this->_helper = $helper;
-    }
-
     /**
      * {@inheritdoc}
      * @SuppressWarnings(PHPMD.ExcessiveMethodLength)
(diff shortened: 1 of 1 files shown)


centerax:
looks good to me. What do you say @gonzaloebiz ?

gonzaloebiz:
Looks OK but i need to test it first, i will try this week

BlackIkeEagle:
@gonzaloebiz thanks

installing within an existing magento installation should work just as before, the only change you could experience is if you don't have proper permissions to create the MailChimp folder.

The big change is when you do a clean install of magento2, with the current stable you will not be able to install with the failure described in the commitbody.

gonzaloebiz:
Thanks @BlackIkeEagle i just test it and it's OK

PR #76, opened 2017-08-30
bugfix STOREID null for error: "Data did not match any of the schemas described in anyOf."

elisongit:
If customer subscribe newsletter as a guest, the variable $subscriber in the function beforeSubscribe (class Ebizmarts/MailChimp/Model/Plugin/Subscriber.php) has storeId null. This situation generate error "Data did not match any of the schemas described in anyOf." when module calls api for add subscribe in mailchimp.

We set storeId for resolve issue. For our company policy we modified also composer version because we editing a module third parts.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
   },
   "description": "Connect MailChimp with Magento",
   "type": "magento2-module",
-  "version": "1.0.22",
+  "version": "1.0.23",
   "authors": [
     {
       "name": "Ebizmarts Corp",
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
   },
   "description": "Connect MailChimp with Magento",
   "type": "magento2-module",
-  "version": "1.0.21",
+  "version": "1.0.22",
   "authors": [
     {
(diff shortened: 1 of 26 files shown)


gonzaloebiz:
Please, don't change the version.

composer.json
```diff
@@ -6,7 +6,7 @@
   },
   "description": "Connect MailChimp with Magento",
   "type": "magento2-module",
-  "version": "1.0.21",
+  "version": "1.0.21.1",
```
centerax: please see Gonzalo's comment. remove this change and the pr can be merged.

centerax (changes requested):

elisongit:
Thanks for reply, we add fourth digit in version composer for manage bugfix in third part modules.
Furthermore we installed Magento via composer and to download the updated version of module is necessary we modify the version in composer.json because composer checks version in composer.json to download module.

gonzaloebiz (changes requested):
Please, don't change the version

gonzaloebiz:
One month without news, so i close the PR

PR #80, opened 2017-09-05
Change $this->_helper->__() to __()

rikardwissing:
function $this->_helper->__() does not exist.


gonzaloebiz:
Thanks @rikardwissing for your feedback

PR #96, opened 2017-10-06
bump version on module.xml to allow schema upgrade

marcatos:
On large catalogs, an index on mailchimp_sync_ecommerce table is required.

This is implemented in etup/UpgradeSchema.php:415, but the version on module.xml is outdated causing the module does not upgrade schema.


gonzaloebiz:
The version will change when we release the new version.

Issue #136, opened 2017-11-22
Not all MailChimp lists are showing in Magento admin configuration


lstellway:
## Preconditions

1. Using Magento `2.1.9`
2. mailchimp/mc-magento `1.0.25`
3. Configure plugin - add API key

## Steps to reproduce

1. Create multiple lists in MailChimp account
2. Navigate to MailChimp configuration 
  - Stores > Configuration > MailChimp > MailChimp
3. Try to choose different lists

## Actual and Expected result

I expect to be able to choose from different lists, however only one shows as an available option.

## Additional information

I have create a pull request that fixes this issue.

gonzaloebiz:
Hi @loganstellway 

this is not the expected result.
The list depends on the mailchimp store, once you select the mailchimp store the list can be only one, one mailchimp store has only one list associated.

Best

lstellway:
@gonzaloebiz 
Ah! Got it - thank you

PR #137, opened 2017-11-22
Fixes Issue #136 (Not all MailChimp lists are showing in Magento admin configuration)

lstellway:


gonzaloebiz:
As i said in the issue https://github.com/mailchimp/mc-magento2/issues/136 this fix is not necessary as this is the expected result

PR #148, opened 2017-12-06
Database Sharding Compatibility for Upgrade Data Script

davidverholen:
Currently, the installation fails when databases are already sharded since the UpgradeData script uses the default connection for sales tables.

This Pull request does not change any logic but uses the sales db connection.

### DIFF
--- a/Setup/UpgradeData.php
+++ b/Setup/UpgradeData.php
@@ -23,19 +23,20 @@ class UpgradeData implements UpgradeDataInterface
         if (version_compare($context->getVersion(), '1.0.24') < 0)
         {
             $setup->startSetup();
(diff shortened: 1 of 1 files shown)


davidverholen:
@gonzaloebiz your kidding right?

For such a small change where I'm doing your work?
besides this is a pull request with a solution provided and not an issue

gonzaloebiz:
Please read the labels description in the home page

davidverholen:
again: this is not an issue, it's a pull request.

an issue is a bug report. A pull request is a bug fix. Without this fix your module is not usable with the enterprise edition.

You really want to tell me now that you are not going to merge it because of a "rule" for issues, which does not even (according to your readme) apply for prs?

PR #167, opened 2017-12-29
Refactor and error catching of the is_syncing flag update process.

SantiParagarino:


centerax (approved):

ihor-sviziev:
@Santiagoebizmarts adding new column to orders grid might be long process that is blocking orders table for big merchants. As result - some period of time customers will not be able to place their orders. I think it's better to create own table and store all needed information about orders in that table.
Any thoughts?

Issue #164, opened 2017-12-27
script/get url path


Detzler:
## URL path

Not able to fetch Mailchimp JS when Magento runs in a subfolder.
The mailchimpjs.js contains the following:

```
require(["jquery"], function ($) {
    $.ajax({
        url: '/mailchimp/script/get',
        type: 'POST',
        dataType: 'json',
        showLoader: false
    }).done(function (data) {
        var imported = document.createElement('script');
        imported.src = data.url;
        document.head.appendChild(imported);

    });
});
```

The url doesn't make usage of the base url parameter

`url: '/mailchimp/script/get',`

### Preconditions

    1. Magento CE 2.2.2 without sample data is installed.
    2. mc-magento2 1.0.25.

### Steps to reproduce

    1. Open any page.

### Actual and Expected result

    Expected result:
    `url: '/my-shop-sub-folder/mailchimp/script/get',`
The url should contain the  base path.

    Actual result:
    `url: '/mailchimp/script/get',`

lstellway:
I opened a pull request ([#172](https://github.com/mailchimp/mc-magento2/pull/172)) to fix the issue.

This is essentially reverting back to the change made in [issues/51](https://github.com/mailchimp/mc-magento2/issues/51), except that now the script will be included in the `head` section. 

Is there a reason this approach wouldn't work?

Detzler:
hey @loganstellway, thanks for your effort but this PR is not satisfiable. I wrote some lines at the PR:
https://github.com/mailchimp/mc-magento2/pull/172#pullrequestreview-92160155

maderlock:
Just a reminder that this should be a GET call, not POST. I've attempted to resurrect the ticket where this was supposed to have been changed.

gonzaloebiz:
Take a look at this https://github.com/mailchimp/mc-magento2/issues/112

PR #172, opened 2018-01-04
Fix for Issue #164

lstellway:
Fixes [issues/164](https://github.com/mailchimp/mc-magento2/issues/164) by removing `mailchimpjs.js` script include and adding `templates/mailchimpjs.phtml` to `head` section.


Detzler (changes requested):
The way the script was included in head was correct. You just need to pass the parameters to the js file.

Please take a look here how it's need to be done:
https://mage2.pro/t/topic/146
https://magento.stackexchange.com/questions/89187/in-magento2-what-is-script-type-text-x-magento-init
https://alanstorm.com/magento_2_javascript_init_scripts/

I dont recommend the solution via head.additional because it bypasses Magento's JS management like merging.

gonzaloebiz:
Please look at the comment by @Detzler

PR #201, opened 2018-02-08
Show "Mailchimp" customer tab when the extension is enabled

### LINKED ISSUE #200: Unable to edit customer without API key
## Preconditions

  1. PHP 7.1.11
  2. Magento CE 2.2.1 without sample data
  3. mc-magento2 1.0.27
  4. mc-magento2 module enabled but _not_ configured (no API key entered, no store connected, etc.)
  5. At least one customer account

## Steps to reproduce

  1. In the admin panel, visit Customers > All Customers
  2. Select a customer and then select the "Edit" option.

##  Expected result:

Customer edit tabs should be shown like normal (account information, addresses, etc.).

## Actual result:

No tabs are shown. An error message is displayed, clobbering nearly the entire edit interface. The error displayed is:

>API Key Invalid : Your API key may be invalid, or you've attempted to access the wrong datacenter.

## Additional information

screenshot from 2018-02-08 16-44-07

t-richards:
This PR fixes the page layout for Customer > Index > Edit so that the "Mailchimp" tab is only displayed when the extension is enabled.

As a side effect, attempting to load the customer's interests no longer happens, which should fix #200 .

### DIFF
--- a/view/adminhtml/layout/customer_index_edit.xml
+++ b/view/adminhtml/layout/customer_index_edit.xml
@@ -15,7 +15,7 @@
     <body>
         <referenceContainer name="content">
             <referenceBlock name="customer_form">
-                <block class="Ebizmarts\MailChimp\Block\Adminhtml\Customer\Edit\Tabs\Mailchimp" name="customer_edit_tab_mailchimp" template="tab/view.phtml">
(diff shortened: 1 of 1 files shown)


gonzaloebiz:
Hi @t-richards 

first of all, thanks for your feedback.
The main idea is OK, the tab must be displayed only when the extension is enabled and your PR fix it.
But i must say that i can't replicate the issue https://github.com/mailchimp/mc-magento2/issues/200
So, you can please change the subject of the PR, that way we maintain documented your PR

Best

t-richards:
The PR title has been changed.

gonzaloebiz (approved):

Issue #202, opened 2018-02-09
When modify a simple product, the variant is empty


gonzaloebiz:
- Modify any simple product data, p.e. the price
- save it

When the product is synced, no data for the variant is sync

PR #226, opened 2018-03-09
Update mc-magento2/Model/Api/Product.php

### LINKED ISSUE #227: Product update in magento admin is not passed to mailchimp
## Issue Description
Product updates are not sent to mailchimp. Changes to price and quantity is not sent after the first sync to mailchimp.

### Steps to reproduce

    1. Create a product
    2. run cron for ebizmarts_ecommerce
    3. Check product data in mailchimp (using the api)
    4. update price or quantity for product inside magento admin.
    5. run cron for ebizmarts_ecommerce again
    6. check product data in mailchimp again. The price and quantity has not changed.

    Expected result:
    Price and quantity should be updated in mailchimp when changed in magento
    Actual result:
    Price and quantity is never changed after the first sync to mailchimp

### Possible solution
$parentIds on line https://github.com/mailchimp/mc-magento2/blob/a14f55a87fe3bad84415f908d5e4b341229cc887/Model/Api/Product.php#L279 
are empty for products with TYPE_SIMPLE. Change the if statement on line https://github.com/mailchimp/mc-magento2/blob/a14f55a87fe3bad84415f908d5e4b341229cc887/Model/Api/Product.php#L269
and add code $variantProducts[] = $product; only for products with type=TYPE_SIMPLE.

rogersv:
Return of $operations was missing in the loop. 
Change of price,qty etc in magento admin was not sent to mailchimp because of this bug.
Everything added inside the loop was overwritten in the end of the function.
Fixes #227

### DIFF
--- a/Model/Api/Product.php
+++ b/Model/Api/Product.php
@@ -307,6 +307,11 @@ class Product
                     $operations[] = $productdata;
                 }
             }
+            
+            if (count($operations) > 0) {
+                return $operations;
+            }
+            
         } elseif($product->getTypeId() == \Magento\ConfigurableProduct\Model\Product\Type\Configurable::TYPE_CODE) {
             $childProducts = $product->getTypeInstance()->getChildrenIds($product->getId());
             $variantProducts[] = $product;


gonzaloebiz:
Hi @rogersv 

your solution is not OK
See at the line https://github.com/rogersv/mc-magento2/blob/9f9f649c4eee4cd077a301d6460d0cb00b4135b5/Model/Api/Product.php#L341
In the loop, the variants are processed and outside the loop the main product, you can't return without processing the main product

rogersv:
Hi @gonzaloebiz,
You are right it was not a solution. I will check the code again. 
I think the real issue is that $variantProducts is empty for products without $parentIds. 
I will check the code again and update issue #227 if I find the cause.

gonzaloebiz:
take a look at this https://github.com/mailchimp/mc-magento2/issues/202

rogersv:
Thanks, I did not see that it was already solved in #202. I added a possible solution in issue #227 but that issue can be closed.

Issue #232, opened 2018-03-13
MailChimp js file loaded each time the page loads


SantiParagarino:
Possible solution: Verify if the config exists before doing the API call [HERE](https://github.com/mailchimp/mc-magento2/blob/1.0.27/Helper/Data.php#L782).

gonzaloebiz:
Closed by https://github.com/mailchimp/mc-magento2/pull/236

PR #236, opened 2018-03-14
Fix getJsUrl method to use config cache properly

wojtekn:
It's a fix for #232, thanks @toshz


gonzaloebiz (approved):
Thanks @wojtekn for your contribution

PR #265, opened 2018-03-28
fix decodeArrayFieldValue error

gundamkid:
Magento 2.2.2
mc-magento: 1.0.28

Argument 1 passed to Ebizmarts\MailChimp\Helper\VarsMap::decodeArrayFieldValue() must be of the type array, string given, called in /var/www/app/code/Ebizmarts/MailChimp/Helper/VarsMap.php on line 48 and defined in /var/www/app/code/Ebizmarts/MailChimp/Helper/VarsMap.php:73

This issue appear when I run setup:upgrade (for old versions too)


gonzaloebiz (approved):
Thanks for your contribution

PR #283, opened 2018-04-17
Add check before renaming 'address_address1' table column

### LINKED ISSUE #235: column "address_address1" does not exist in table "mailchimp_store"
I used the composer to install it，
then ,"php bin/magento setup:upgrade"，The above problems arise

ghost:
## Details

* Add conditional `if` check before performing the table column rename
* Fixes #235 

### Issue description
----------------------
When installing version `1.0.28` of Mailchimp for M2 via composer and running `bin/magento setup:upgrade` the following error *sometimes* occurs as outlined in issue #235 :

```
Column "address_address1" does not exist in table "mailchimp_stores".
```

This prevents the setup process from completing and therefore neither the storefront nor the backend will work due to invalid module versions from the `setup_module` table. 

### What *should* happen?
-----------------------------
The setup process should complete as normal and storefront/backend should work.

### Fix
------------
Upon checking the `mailchimp_stores` table, the columns have already been renamed to `address_address_one` and `address_address_two`, so the issue is because this function is being triggered when it shouldn't be. While I am not suggesting this is a permanent fix, wrapping the table rename function with a conditional statement will alleviate the issue until a resolution is found.

### DIFF
--- a/Setup/UpgradeSchema.php
+++ b/Setup/UpgradeSchema.php
@@ -296,28 +296,32 @@ class UpgradeSchema implements UpgradeSchemaInterface
             );
         }
         if (version_compare($context->getVersion(), '1.0.12') < 0) {
-            $connection->changecolumn(
-                $setup->getTable('mailchimp_stores'),
-                'address_address1',
-                'address_address_one',
-                [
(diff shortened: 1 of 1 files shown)


gonzaloebiz (approved):
Hi @RobAimes 

thanks for your contribution.

Best

Issue #345, opened 2018-07-11
Unable to set custom Env.php API Credentials 


RidRack:
### Title
I'm unable to set the API credentials using the Env.php file as there is an exception in the file https://github.com/mailchimp/mc-magento2/blob/develop/Model/Config/Backend/ApiKey.php on line `81`. 

The exception is:
```
[ ok ] Starting enhanced syslogd: rsyslogd.
Processing configurations data from configuration file...
Import failed: Notice: Undefined variable: active in /var/www/magento/vendor/mailchimp/mc-magento2/Model/Config/Backend/ApiKey.php on line 81
```

This is caused by the `active` variable not being assigned before being used. 

### Preconditions
Simply set an API key in the env.php file like follows:
```
 /**
             * Mailchimp Settings
             */
            'mailchimp' => [
                'general'       => [
                    'active'     => 0,
                    'apikeylist' => 'xxxxxxxxxxxxxxxxxxxxxxxx-us15',     
                ],
            ],
```
and then run the magento `app:config:import` command.

### Steps to reproduce
1. Add mailchimp env.php values, setting an API key
2. run the `app:config:import` command

### Actual and Expected result
An exception is thrown:
```
Processing configurations data from configuration file...
Import failed: Notice: Undefined variable: active in /var/www/magento/vendor/mailchimp/mc-magento2/Model/Config/Backend/ApiKey.php on line 81
```

PR #346, opened 2018-07-11
Issue345: Ensure variable active is assigned

RidRack:
Ensures that the variable active is assigned before it's used, allowing the empty check later to work correctly and config be set using env.php

Resolves https://github.com/mailchimp/mc-magento2/issues/345


gonzaloebiz (approved):

gonzaloebiz:
Thanks for your contribution

PR #377, opened 2018-08-27
Adding extra logging on import parts of the Cronjob

maximbaibakov:
Adding extra logging on import parts of the Cronjob.

There is a bit lack of visibility around execution of the sync process.
Would be awesome to see where the sync spends more time and helps to diagnose, if the sync stuck somewhere due to some reason.

E.g. Reasons:
- Payload was not able to generate due to M2 Core issue and failed with warning or fatal error
- You can see how much time been spent on specific part e.g. products, orders and etc. It helped me to diagnose that cart generation was taking about 20 minutes to generate due to some slow SQL query

### DIFF
--- a/Cron/Ecommerce.php
+++ b/Cron/Ecommerce.php
@@ -170,19 +170,30 @@ class Ecommerce
         $batchArray = [];
         $results = $this->_apiSubscribers->sendSubscribers($storeId, $listId);
         if ($this->_helper->getConfigValue(\Ebizmarts\MailChimp\Helper\Data::XML_PATH_ECOMMERCE_ACTIVE, $storeId)) {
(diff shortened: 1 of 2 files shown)


gonzaloebiz (changes requested):
Please open an issue for the extra logging issue and delete the extra commit for the temp directory (this is in a separate issue and already accepted)

maximbaibakov:
Hi @gonzaloebiz 

Thanks for the review. Issue has been created: https://github.com/mailchimp/mc-magento2/issues/393

PR has been adjusted

Regards,
Maxim

gonzaloebiz (approved):

gonzaloebiz:
Hi @maximbaibakov 

thanks for your contribution

Issue #378, opened 2018-08-27
MailChimp breaks order processing when it's done through CLI


wojtekn:
### Preconditions

The issue exists with any Magento 2 version and latest mc-magento2 version.

### Steps to reproduce

It's not possible to create steps to reproduce using default Magento with sample data as it doesn't include any code which processes orders from CLI. 

The issue appears when  there is used a custom module which processes orders using CLI - eg. subscription module which creates orders using saved token payment method on CRON basis.

It's caused by \Ebizmarts\MailChimp\Observer\Sales\Order\SubmitAfter::execute code running in all scopes.

### Actual and Expected result

    Expected result:
    Order is placed successfully by CLI.
    Actual result:
    Script terminates with the error.

gonzaloebiz:
Thanks for your feedback,
Please next time put the closes comment in the commit that closes the issue

Best

PR #379, opened 2018-08-27
Move observer to frontend scope to avoid breaking CLI orders

wojtekn:
This is issue #378.

PR moves mailchimp_sales_model_service_quote_submit_after observer to frontend scope to avoid throwing cookie error on CLI.


gonzaloebiz (approved):
closes #378

Issue #388, opened 2018-09-11
Failed to open stream: No such file or directory


maximbaibakov:
## Issue Description

- MailChimp.log generates errors
- mailchimp_sync_batches batch table says that batch pending or cancelled.

### Preconditions

    1. Magento CE 2.2.3 without sample data is installed.
    2. mc-magento2 1.0.29

### Steps to reproduce

1. Install Magento CE 2.2.3
2. Install mc-magento2 1.0.29
3. Create order
4. Sync data with Mailchimp

### Expected result:

- Data is synced
- No errors in log
- mailchimp_sync_batches has updated and confirmed that batch was successful

### Actual result:

- Data is synced
- MailChimp.log generates errors 
- [2018-09-11 23:05:33] MailChimpLogger.INFO: Warning: fopen(/.../var/Mailchimp/c5b6315033.tar.gz): failed to open stream: No such file or directory in /.../vendor/mailchimp/mc-magento2/Model/Api/Result.php on line 101
- mailchimp_sync_batches batch table says that batch pending or cancelled.

maximbaibakov:
Hi there,

PR has been opened to resolve the issue https://github.com/mailchimp/mc-magento2/pull/389

Regards,
Maxim

gonzaloebiz:
Hi @maximbaibakov  thanks for your contribution

PR #389, opened 2018-09-12
Create temporary directory, if that does not exist

maximbaibakov:
Create temporary directory, if that does not exist to avoid error "Failed to open stream: No such file or directory"


gonzaloebiz:
Hi @maximbaibakov 

thanks for your contribution, can you please make 2 different PR, one for the fix of the directory and other for the logging?

Thanks

gonzaloebiz (approved):
closes #388

Issue #400, opened 2018-09-27
Ecommerce cronjob stuck on customers with no address (error occurs)


stephan-cream:
### Preconditions
1. Any Magento version.
2. Any server setup
3. Customer without an address
4. Ecommerce cronjob active

### Steps to reproduce
1. Create a customer (frontend / backend both experience this).
2. Do not add an address to the customer.
3. Add 'Default Billing Address' or 'Default Shipping Address' to the configuration 'Customer Fields Mapping' in "Stores => Configuration => Mailchimp => Mailchimp General Configuration".
4. Allow the cronjob to run (or use magerun for manual execution).
5. Check the `cron_schedule` table in the database.

### Expected result
1. The customer is skipped altogether during the cronjob OR the customer is synchronised without an address.
2. The cronjob is executed fully without any exceptions occuring.

### Actual result
1. The customer is not skipped and an address is expected.
2. An error is thrown: `[2018-09-04 09:05:06] main.CRITICAL: Error when running a cron job {"exception":"[object] (RuntimeException(code: 0): Error when running a cron job at /vendor/magento/module-cron/Observer/ProcessCronQueueObserver.php:324, TypeError(code: 0): Argument 1 passed to Ebizmarts\\MailChimp\\Helper
Data::_getAddressValues() must be an instance of Magento\\Customer\\Model\\Address
AbstractAddress, boolean given, called in /vendor/mailchimp/mc-magento2/Helper/Data.php on line 563 at /vendor/mailchimp/mc-magento2/Helper/Data.php:583)"} []`
3. The cronjob stops execution and never synchronises any other customers as it keeps failing on the customer that had no address.

gonzaloebiz:
Hi @stephan-cream 

thanks for your feedback.
I can't reproduce your issue, i follow your steps and the customer was synced fine.
Magento version 2.2.5 with sample data
mc-magento2 version 1.0.30
You have any other extensions installed? Try with a vanilla installation

Best

stephan-cream:
My apologies, i forgot a step.
**After step 2:** Add 'Default Billing Address' or 'Default Shipping Address' to the configuration 'Customer Fields Mapping' in "Stores => Configuration => Mailchimp => Mailchimp General Configuration".
This is when the error occurs during the cronjob.
Although you could say it is required to have an address when this is set, it is obviously impossible to force this and it should not block the cronjob from synching the rest.

I updated the initial comment.

gonzaloebiz:
[...]

PR #401, opened 2018-09-28
Customer with no address issue

arendarenko:
My PR contains two bug-fixes:
- Fix for #400 
- There is no created issue for that case: If there a products with no price, Mailchimp returns an while syncing error: "variants.item:0.price Schema describes number, NULL found instead"


gonzaloebiz:
Hi @alexeya-ven 

thanks for your contribution.
Please make 2 PR, one per issue (open an issue for the price issue). 
This way we can generate automatically the changelog, and any other user can find if the issue was solved. If you put more than 1 issue in the same PR, the users only view the one related to the created issue and the other remains hidden for the most of the users.

Best

PR #403, opened 2018-09-28
Fix bug with NULL price when special price is not set, but special price dates are set

arendarenko:
Fix for #416

### DIFF
--- a/Model/Api/Product.php
+++ b/Model/Api/Product.php
@@ -281,7 +281,7 @@ class Product
                     $variendata["title"] = $data["title"];
                     $variendata["url"] = $data["url"];
                     $variendata["sku"] = $data["sku"];
-                    $variendata["price"] = (float)$data["price"];
+                    $variendata["price"] = $data["price"];
                     $variendata["inventory_quantity"] = $data["inventory_quantity"];
                     $variendata["image_url"] = $data["image_url"];
                     $variendata["backorders"] = $data["backorders"];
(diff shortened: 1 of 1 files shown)


gonzaloebiz:
Hi @alexeya-ven 

thanks for you feedback.
What do you mean with "products with no price"? The price is a required field, in fact i can't create a product in the backend without price 

image

arendarenko:
Hi @gonzaloebiz 
I've updated my PR.
You're right, by default u cant set product with 0 price from backend.
But I've had an errors like this: "variants.item:0.price : Schema describes number, NULL found instead" and I started investigate why they are happening. 

Steps to reproduce:

1. Create product with price $40
2. Set such advanced price options:
2.1. Special price = empty field
2.2. Special price from = any past date (for example: 30/09/2018)
3. Save & try to import it into MailChimp

Expected result:
1. Product will be exported with price $40

Actual result:
1. Product will try to exported with NULL price (because module tries to use special price if it's date range is correct, but didn't check exact value of advanced price!). As result, product will NOT exported to MailChimp and you will receive an error: "variants.item:0.price Schema describes number, NULL found instead".

gonzaloebiz:
Hi @alexeya-ven 

thanks for your contribution.
The fix is ok, but we need that your create the corresponding issue as described in our [pull request guidelines](https://github.com/mailchimp/mc-magento2/wiki/Pull-Request-guidelinessl). This is for keep the repo organized in benefit of our users and allow to us to and others to know what issue was fixed, when was fixed, in which version the fix is included, who fixed the bug etc.
p.e. the subject of this pr is wrong, the pr not fix this situation

Best

arendarenko:
Hi @gonzaloebiz 
I've created an issue #416, please check is everything is correct

gonzaloebiz (approved):
Thanks @alexeya-ven for your contribution

Issue #417, opened 2018-10-05
Incorrect re-syncing modified subscribers


arendarenko:
## Issue description

When you're modifying subscriber status from Magento data is not sending to Mailchimp.

### Preconditions:
1. Installed Magento v2.2.5
2. mc-magento  v1.0.30

## Steps to reproduce

1. Create new customer in Magento, subscribe to newsletter
2. Wait untill subscriber will be appear in Mailchimp list
3. Unsubscribe this customer from newsletter

### Expected result
1. Subscriber will be unsubscribed in Mailchimp

### Actual result
1. Subscriber is still in "Subscribed" status in Mailchimp

PR #418, opened 2018-10-05
Fix subscribers update status

arendarenko:
Fix for #417 

This changes makes possible to update subscriber status from storefront after adding it to Mailchimp.


arendarenko:
It doesn't work well. Subscriber status updates in Mailchimp, but it also affects subscriber status in Magento when syncing with Mailchimp. So when user unsubscribe himself in Magento, when syncing he unsubscribes in MailChimp and THEN user automaticly subscribes in Magento (possibly because of webhooks).

PR #450, opened 2018-10-31
Fix unique type validation in di.xml

ihor-sviziev:
This PR fixes unique type validation errors in di.xml file.

**Steps to reproduce:**
1. Install magento 2.2.6 + latest mailchimp module
2. Open PhpStorm and add URN highliter - Howto: https://devdocs.magento.com/guides/v2.2/config-guide/cli/config-cli-subcommands-urn.html
3. Open `vendor/mailchimp/mc-magento2/etc/di.xml` file and see if file validation is success (no underlines near `<"type"... >` lines)


gonzaloebiz (approved):
Thanks for your contribution

Issue #459, opened 2018-11-06
Error: MailChimpLogger.INFO: Resource Not Found for Api Call: https://us11.api.mailchimp.com/3.0/lists//interest-categories using method [GET]


arendarenko:
### Preconditions

Example:

    1. Magento CE 2.2.6 without sample data is installed.
    2. mc-magento2 1.0.30.
    3. Interests are not set
    ...

### Steps to reproduce

1. Setup mailchimp module without setting interests settings
2. Add new subscribers in Magento2
3. Run module cronjobs

### Expected result:

1. Subscribers will be sent to MailChimp successfully without any error

### Actual result:

1. Subscribers are added but we also receiving error in Mailchimp.log:
`[2018-11-06 09:48:08] MailChimpLogger.INFO: Resource Not Found for Api Call: https://us11.api.mailchimp.com/3.0/lists//interest-categories using method [GET]`

duckchip:
Hi there!
Is there any progress on this issue? :)

gonzaloebiz:
Hi @duckchip 

please read at https://github.com/mailchimp/mc-magento2/pull/460

Best

pmathbliss:
#459 rereferences this issue.

jaimiedijstra:
Just come across this error on my site too, Magento 2.3.5p2 and Mailchimp v102.3.41. Just wanted to check this here with the devs as this issue loop references issue #460 so no clear answer.
On my cron time and logs this issue just crops up now and again (only twice in the last 24 hours, no load issues on my server or other issues), but then works fine again, is this just connectivity Mailchimp side and no concern?
![image](https://user-images.githubusercontent.com/12552943/100606065-cff99d80-3343-11eb-8d59-6d50e6893db4.png)

PR #460, opened 2018-11-06
Fix interest empty list issue

arendarenko:
Fir for #459 459

### DIFF
--- a/Cron/Ecommerce.php
+++ b/Cron/Ecommerce.php
@@ -169,7 +169,9 @@ class Ecommerce
         $countOrders = 0;
         $batchArray = [];
         $this->_helper->resetCounters();
-        $results = $this->_apiSubscribers->sendSubscribers($storeId, $listId);
+        if ($listId) {
+            $results = $this->_apiSubscribers->sendSubscribers($storeId, $listId);
+        }
         if ($this->_helper->getConfigValue(\Ebizmarts\MailChimp\Helper\Data::XML_PATH_ECOMMERCE_ACTIVE, $storeId)) {
             $this->_helper->log('Generate Products payload');
             $products = $this->_apiProduct->_sendProducts($storeId);


gonzaloebiz (changes requested):
This change has nothing related to the interest groups, listId contains the mailchimp list associated to the mailchimp store an has no relation with the interest groups

arendarenko:
hi @gonzaloebiz 
sendSubscribers calls getInterest method: https://github.com/mailchimp/mc-magento2/blob/develop/Model/Api/Subscriber.php#L60 which tries to retrieve listId by storeId https://github.com/mailchimp/mc-magento2/blob/develop/Helper/Data.php#L1003 without any check (for example, if listId is not empty, API method will executed this way: "lists//interest-categories", which will cause this error.

gonzaloebiz (commented):
well, on there https://github.com/mailchimp/mc-magento2/blob/develop/Cron/Ecommerce.php#L125 we check if the extension is active, then o there https://github.com/mailchimp/mc-magento2/blob/develop/Cron/Ecommerce.php#L140 we check if you select a mailchimp store.
Each mailchimp store has a list, no way to define a mailchimp store without a list, so you can't have a mailchimp store configured but no list. 
I think the only way to have no list is modify the database by hand

gonzaloebiz:
No news in 22 days

ihor-sviziev:
Hi @gonzaloebiz ,

We have following configuration:
- Default scope:
capture 2019-01-04 at 10 02 53
    - Website 1 (configured for MC store ID: "Website 1", list: "Website 1")
capture 2019-01-04 at 10 04 54
    - Website 2 (configured for MC store ID: "Website 2", list: "Website 2")
capture 2019-01-04 at 10 07 22
    - Website 3 - not configured
capture 2019-01-04 at 10 08 18

Each website has one store view, nothing configured on store view level.

In such configuration looks like for the store related to website 3 somehow we're getting into this issue.
Could you re-check it?

PR #470, opened 2018-11-12
Prevent "Unable to unserialize value" exception

flancer64:
"Unable to unserialize value" exception is thrown in case of `$interestGroup->getGroupdata()` has no data.

I've just added this validation:
```
        $serialized = $interestGroup->getGroupdata();
        if($serialized) {
            $groups = $this->_serializer->unserialize($serialized);
            ...
        }
```


gonzaloebiz:
Hi @flancer64 

thanks for you contribution, we made a more strong check, anyway really thanks for your feedback.

Best

Issue #545, opened 2018-12-22
Ecommerce cron job fails when customer has no interest group data


aligent-lturner:
## Issue Description
When at least one customer exists with no interest group data, the Ecommerce cron job fails with the following error:
```
Error: Unable to unserialize value.
```

### Preconditions

    1. Magento EE 2.2.2 (though will happen for any version)
    2. mc-magento2 1.0.31.

### Steps to reproduce

    1. Create two customer accounts via front-end or admin.
    2. Ensure one customer has no interest group record in the database
    3. Run Ecommerce cron job.

### Actual and Expected result

    Expected result:
    Job completes successfully
    Actual result:
    Job fails with error: `Error: Unable to unserialize value.`

### Additional information

The error occurs due to an attempt to unserialize a null value. This was introduced by the commit 
https://github.com/mailchimp/mc-magento2/commit/a6159df0b5e1e83a3fb334e1aa928af5eef19b81
The Magento unserialize function will throw an exception, whereas the default PHP unserialize will just return false.

gonzaloebiz:
This issue is already fixed

PR #546, opened 2018-12-22
Prevent exception being thrown when unserializing null value

aligent-lturner:
This change is a fix for #545 - the Magento unserialize function throws an exception when attempting to unserialize a null value. This change replaces null values with empty JSON data to prevent the exception from being thrown and causing the Ecommerce cron job to fall.


gonzaloebiz:
already fixed

PR #590, opened 2019-02-12
Save email to quote configuration

ghost:
Saving customer's email to the quote without agreement from the customer leads to GDPR non-compliance. By this pull request, I'm going to add a store level configuration about saving email to the quote. By default, it is enabled.  if you want to be GDPR compliant you should disable saving customer identity information to the database without agreement from the customer.


gonzaloebiz:
Sorry but i'm not sure about this, we are asking to the legal staff.
By the way, what about the order or even registered customers? Because magento save the email by default the email in the quote for registered customers, and of course save the email in the order for all, registered and guest customers.

Best

gonzaloebiz:
Hi @olegAleksin 

thanks for your feedback.
We open a new issue for this (https://github.com/mailchimp/mc-magento2/issues/645) because we need to fix it for m2.1 m2.2 and m2.3

Best

PR #856, opened 2019-08-27
Made sure a missing "simple_sku" won't break the entire sync.

sicet7:
Made sure that a missing "simple_sku" from the productOptions wont break the entire sync.


gonzaloebiz:
Please take a look at https://github.com/mailchimp/mc-magento2/issues/862

PR #867, opened 2019-09-04
Fix catching campaign with enabled Varnish FPC

ihor-sviziev:
### Description
After applying changes from https://github.com/magento/magento2/pull/21818 - I got an issue - campaign IDs is always empty when customer went from Mailchimp.

Motivation for this fix was following - all marketing get params should be removed on Varnish in order to improve using Full Page Cache.

In order to fix this issue - we're sending `mc_cid` not as get parameter `mailchimp/campaign/check/?mc_cid=<campaign_id>`, but as controller get params `mailchimp/campaign/check/mc_cid/<campaign_id>/`.

Also if request with mc_cid get param was sent to homepage, category page, product page - window.FORM_KEY is `undefined`, so it's not even sent to backend. Also it's not required in controller, so I just removed sending form key.

### Pre-requisites
1. Magento OS 2.3.2
2. Mailchimp module v102.3.36
3. Configured FPC in Varnish mode

### Steps to reproduce
1. Go to following page few times:  https://magento2.dev/?mc_cid=any_valid_campaign_id&
2. See logs in nginx/apache server

### Expected result
1. In nginx/apache logs you should see that  to `/mailchimp/campaign/check/?mc_cid=any_valid_campaign_id` or `/mailchimp/campaign/check/mc_cid/any_valid_campaign_id/`
2. Responce should be `{"valid":1}` all the time
3. This request should not contain dynamic part `_=<current_timestamp>`

Nice to have (*not implemented*): such requests should be cached in Varnish (FPC), so when customers will go from the same campaign - result will be returned from cache.

### Actual result
1. In nginx/apache logs you can see requests like `/mailchimp/campaign/check/?_=1567593457848` and `/mailchimp/campaign/check/?_=1567593886531` (mc_cid was removed in Varnish)
2. Responce is `{"valid":0}` all the time
3. This request contains dynamic part `_=<current_timestamp>`

### DIFF
--- a/view/frontend/web/js/campaigncatcher.js
+++ b/view/frontend/web/js/campaigncatcher.js
@@ -64,10 +64,10 @@ define(
                     }
                     if (mc_cid && !isMailchimp) {
                         $.ajax({
-                            url: checkCampaignUrl,
-                            data: {'form_key': window.FORM_KEY,'mc_cid': mc_cid},
+                            url: checkCampaignUrl + 'mc_cid/' + mc_cid + '/',
                             type: 'GET',
                             dataType: 'json',
+                            cache: true,
                             showLoader: false
                         }).done(function (data) {
                             if (data.valid==0) {


arendarenko (approved):

ihor-sviziev:
Hi @gonzaloebiz,
Could you review my PR?

gonzaloebiz:
Hi @ihor-sviziev thanks for your contribution!!
Sorry for the delay, we are testing the new version and we have busy days
I'm open a new issue with this (https://github.com/mailchimp/mc-magento2/issues/874) to fix it for all the magento versions (and fix a little issue with the js). So i will close this pr, but your fix works and will be included in the next version.

PR #891, opened 2019-10-10
Optimize MailChimp JS block

andrey-legayev:
Avoid initialization of all classes via Magento DI if JS URL has been already cached.
It saves 3-7ms on every shopping cart page load.

### DIFF
--- a/Block/Mailchimpjs.php
+++ b/Block/Mailchimpjs.php
@@ -47,8 +47,14 @@ class Mailchimpjs extends \Magento\Framework\View\Element\Template
     {
         $storeId = $this->_storeManager->getStore()->getId();
 
-        $url    = $this->_scopeConfig->getValue(\Ebizmarts\MailChimp\Helper\Data::XML_MAILCHIMP_JS_URL, ScopeInterface::SCOPE_STORES, $storeId);
-        $active = $this->_scopeConfig->getValue(\Ebizmarts\MailChimp\Helper\Data::XML_PATH_ACTIVE, ScopeInterface::SCOPE_STORES, $storeId);
+        $url    = $this->_scopeConfig->getValue(
+			\Ebizmarts\MailChimp\Helper\Data::XML_MAILCHIMP_JS_URL, ScopeInterface::SCOPE_STORES,
+			$storeId
+		);
+        $active = $this->_scopeConfig->getValue(
+			\Ebizmarts\MailChimp\Helper\Data::XML_PATH_ACTIVE, ScopeInterface::SCOPE_STORES,
+			$storeId
+		);
 
         // if we have URL cached or integration is disabled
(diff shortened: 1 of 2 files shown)


gonzaloebiz (changes requested):
Don't use ObjectManager directly
Don't use the name of a class as a string, use the ::class notation

andrey-legayev:
I've updated pull request: replaced ObjectManager usage by Proxy configuration in di.xml

etc/di.xml
```diff
@@ -11,6 +11,7 @@
  */
 -->
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:ObjectManager/etc/config.xsd">
+    <preference for="Ebizmarts\MailChimp\Helper\Data" type="Ebizmarts\MailChimp\Helper\Data\Proxy" />
```
gonzaloebiz: please change this definition for 
`    <type name="Ebizmarts\MailChimp\Block\Mailchimpjs">
        <arguments>
            <argument name="helper" xsi:type="object">Ebizmarts\MailChimp\Helper\Data\Proxy</argument>
        </arguments>
    </type>
`

this way the change affect only to the Mailchimpjs block

gonzaloebiz (changes requested):

etc/di.xml
andrey-legayev: I did it globally, because I saw usage of the helper in other MailChimp blocks. I think it could also impact performance there (for example on Success page).
But I did not measure performance impact and I didn't get into details of blocks logic.
Let's keep it as you suggested.

Block/Mailchimpjs.php
```diff
@@ -44,6 +46,17 @@ public function __construct(
     public function getJsUrl()
     {
         $storeId = $this->_storeManager->getStore()->getId();
-        return $this->_helper->getJsUrl($storeId);
+
+        $url    = $this->_scopeConfig->getValue(\Ebizmarts\MailChimp\Helper\Data::XML_MAILCHIMP_JS_URL, ScopeInterface::SCOPE_STORES, $storeId);
+        $active = $this->_scopeConfig->getValue(\Ebizmarts\MailChimp\Helper\Data::XML_PATH_ACTIVE, ScopeInterface::SCOPE_STORES, $storeId);
```
gonzaloebiz: these 2 lines exceeds maximum limit of 120 characters

gonzaloebiz (changes requested):

Block/Mailchimpjs.php
andrey-legayev: anything else?

gonzaloebiz (approved):
Thanks @andrey-legayev

PR #893, opened 2019-10-11
Add synchronizable status filter

tdgroot:
We were missing the ability to control which orders are being synchronized to Mailchimp by their statuses, so we added the ability to do that. 

Also, in one case, our client has the Magento_Newsletter module disabled, but uses Mailchimp. Because of that, the `newsletter_subscriber` table does not exist. This resulted in the synchronzation process crashing.

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -856,7 +856,7 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
     }
     public function loadStores()
     {
-        
+
         $mcUserName = [];
         $connection = $this->_mailChimpStores->getResource()->getConnection();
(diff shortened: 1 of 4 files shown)


gonzaloebiz:
Hi @tdgroot 

thanks for choosing our extension.
The first comment, don't include more than one fix in a pr or issue. 
Why do you think that we need to specify order statuses? All the orders are synced, and when the order changes their status is re-synced (the objective here is that mailchimp know when an order was paid or shipped to send the corresponding automation), i think that has no sense to have selected statuses.
If you want to discuss this, please open an issue (a pr is not the way)

Best

PR #907, opened 2019-10-24
[2.3] Group entities by id to prevent duplicate errors

barryvdh:
When 2 entries exists with the same ID, Magento will throw an Exception which errors the cronjob `ebizmarts_ecommerce`

>  Item (Magento\Catalog\Model\Product\Interceptor) with the same ID "xx" already exists.

Grouping by id will only load the entity once.
(Not sure why duplicate entries exist, but I ran into this with both products and customers)


gonzaloebiz:
Hi @barryvdh 

can you please open an issue about following the [issue guidelines](https://github.com/mailchimp/mc-magento2/wiki/Issue-reporting-guidelines)

Best

barryvdh:
What is not clear about the description? I don't think I have more information, it's a M2.3 store with just this error in the cron.

gonzaloebiz:
I need your magento version, your mc-magento2 version, how to replicate the issue (you have all the data in the issue guidelines)

Best

PR #1101, opened 2021-02-09
Update events.xml

thespacersm:
There is an apostrophe in "mailchimp_newsletter_subscriber_save_before" observer name that breaks Magento 2 DOM XML parse.


bigware (changes requested):
Please post this change asap as it crashes magento currently.

SantiParagarino (approved):

Issue #1108, opened 2021-02-23
Error with option "Magento always manage emails" in Mailchimp configuration


jcarlosebizmarts:
Issue Description:

Magento seems to always send the emails of confirmation and welcome ignoring if the "Magento always manage emails" setting is either on yes or no.

Preconditions:

    1. Magento 2.4.2 with sample data installed.
    2. mc-magento2 v102.3.42.
    3. Mailchimp and Magento are configured with Double-opt in.
    4. Mailchimp has the option "Send a final welcome email" checked in. 
    5. Magento's Newsletter has the option "Need to confirm" on.
    6. Mailchimp for Magento has "Magento always manage emails." off.

Steps to reproduce:

    1. Enter the Main Site either as a Guest or a Customer.
    2. Subscribe to Newsletter from Footer Form or Account Settings.
    3. Receive the confirmation email.
    

Actual and Expected result:

    Expected result:
    Suscription and welcoming email are received directly from Mailchimp.
    Actual result:
    Both emails are received from Magento regardless of configuration.    

Additional information

We had Mandrill installed to make these tests, we tested both with "Magento always manage emails" on and off, and enable double-opt in, on and off in both Magento and Mailchimp.

sedax90:
I think this problem could be related (and solved) with this: https://github.com/mailchimp/mc-magento2/issues/1089

gonzaloebiz:
Please take a look at https://github.com/mailchimp/mc-magento2/issues/1089

Issue #1089, opened 2021-01-11
The confirmation email is sent twice because the getImportMode () method cannot be honored


sedax90:
### Preconditions

    1. Magento CE 2.4.1 without sample data is installed.
    2. mc-magento2 102.3.41
    3. Mailchimp double Opt In is activated
    4. Magento Config --> Mailchimp --> Magento always manage emails --> NO
    5. Magento Config --> Customer --> Newsletter --> Confirmation required --> YES

### Steps to reproduce
Subscribe to newsletter.

### Actual and Expected result
You will get two emails instead of once. One from Magento and one from Mailchimp.

### Additional information
I think the problem could be in Plugin beforeSubscribe(). This method add import_mode to true, but when magento subscribe a user it creates a new Subscriber object:

\Magento\Newsletter\Model\SubscriptionManager.php
```
$subscriber = $this->subscriberFactory->create()->loadBySubscriberEmail($email, $websiteId);
```

The correct way could be to add this data in the afterLoadBySubscriberEmail method.
I've create a new Plugin with this code, and now it works correctly:

```
class Subscriber
{
    protected Data $_helper;

    protected LoggerInterface $logger;

    /**
     * Subscriber constructor.
     * @param Data $_helper
     * @param LoggerInterface $logger
     */
    public function __construct(Data $_helper, LoggerInterface $logger)
    {
        $this->_helper = $_helper;
        $this->logger = $logger;
    }

    /**
     * @param \Magento\Newsletter\Model\Subscriber $subscriber
     * @param $result
     * @return mixed
     */
    public function afterLoadBySubscriberEmail(\Magento\Newsletter\Model\Subscriber $subscriber, $result)
    {
        try {
            if (!$this->_helper->getConfigValue(
                Data::XML_MAGENTO_MAIL,
                $result->getStoreId()
            )) {
                $result->setImportMode(true);
            }
        } catch (\Exception $exception) {
            $this->logger->critical($exception->getMessage());
        }

        return $result;
    }
}
```

denis-zyk:
Hi @sedax90,

I was wondering what is the plan on adding the changes mentioned here to a new module release?

artttj:
Hi @gonzaloebiz,

I assume the issue with double opt-in (#1108) is resolved now, isn't it?  
When to expect a new release?

gonzaloebiz:
Hi @artttj 

yes, the issue is solved.
We don't have an ETA for a new release, sorry

Best

[...]

PR #1119, opened 2021-03-03
Patch for double confirmation mail

sedax90:
Resolve issues like: https://github.com/mailchimp/mc-magento2/issues/1089 and https://github.com/mailchimp/mc-magento2/issues/1108


gonzaloebiz:
Hi @sedax90 

thanks for your contribution.
The fix is a bit more complicated than this (see at https://github.com/mailchimp/mc-magento2/pull/1120), but your contribution helps me, so thanks

Best

PR #1125, opened 2021-03-05
closes #1107 for magento 2.4

### LINKED ISSUE #1107: Error table never getting cleaned up
# Issue reporting guidelines

To maintain an effective bugfix workflow and make sure issues will be solved in a timely manner we kindly ask reporters to follow some simple guidelines.

Before creating an issue, please do the following:

* Make sure the behavior you are reporting is really a bug, not a feature.
* Check the existing [issues](https://github.com/mailchimp/mc-magento2/issues) to make sure you are not duplicating somebody’s work.
* Make sure, that information you are about to report is a technical issue, please refer to the [Community Forums](http://ebizmarts.com/mailchimp-for-magento-support)  for technical questions.

If you are sure that the problem you are experiencing is caused by a bug, file a new issue in a Github issue tracker following the recommendations below.

## Title

Title is a vital part of bug report for developer and trigger to quickly identify a unique issue. A well written title should contain a clear, brief explanation of the issue, making emphasis on the most important points.

Good example would be:

> Unable to place order with Virtual product and PayPal.

Unclear example:

> Can't checkout.

## Issue Description
The mailchimp_error table is never cleaned up resulting in 1.5G of physical space used on the disk for this table alone.
I'm also seeing error logs for batches that do not exist in the database anymore

### Preconditions

Magento 2.4.1
mc-magento2 102.3.41

### Steps to reproduce

Let some batches fail and log to the mailchimp_error table
Let these batches expire and get cleaned up by the mailchimp crons

The error logs of these expired batches will still exist in the mailchimp_errors table and are never cleared.

I expect these to be deleted with the batches they are linked to, or to expire themselves after x amount of time

### Additional information

As a possible solution we could add a Mysql relationship between the batches and the errors with `ON DELETE CASCADE` to automatically remove these logs when deleting batches.
Or add the removal of the corresponding errors in the cleanup crons.

gonzaloebiz:


SantiParagarino (approved):

barryvdh:
This query seems not optimal. It seems to keep hanging because of the number of records. Maybe this can be optimized by using a join instead of a subquery? Or better, make the batch_id a foreign key?

PR #1178, opened 2021-05-17
Update ImportAfter.php

extroniks:
Fixes following issue:
1. When running product import, when store view code is missing from the import sheet the ImportAfter observer fails due to undefined index "_store"

Proposed fix: checks if the _store index is set, otherwise passes NULL.


gonzaloebiz:
See issue https://github.com/mailchimp/mc-magento2/issues/1224

Issue #1242, opened 2021-07-09
Deleting error records causes database lockup


barryvdh:
## Title
Deleting error records causes database lockup

## Issue Description

The mailchimp_error table is very big in our case (around 2 million records with recent errors). The deleting in the cron never finishes. This was especially a problem when the fix was applied to actually delete the records (see https://github.com/mailchimp/mc-magento2/issues/1107), with over 20 million records needed to be deleted. This caused a lockup for the entire table. But still with just the 2 million records, the query keeps running and locking up the cron.

### Preconditions

Magento 2.4.2-p1 with latest version of this module. And a lot of history (>2 million errors)

### Steps to reproduce

Running the query doesn't actually finish (at least not in a couple of hours). Just run the cron or the query manually and you will see.

### Actual and Expected result

Expected is that items are deleted. Instead it never finishes and locks the table.

### Additional information

- the batch_id doesn't seem to have an index. I think this could be a foreign key with a cascade. If the batches are deleted, the errors should cascade down. This removes the need for the extra query. Otherwise a normal index might help in preventing a table lock.
- Instead of using 2 subselects, maybe the query can be rewritten to use a left join, to improve performance.

PR #1243, opened 2021-07-14
Optimize batch cleanup

barryvdh:
This optimizes the batch cleanup:

1. First delete the inactive batches
2. Then delete only the errors without batches (because you already deleted the inactive ones before now)
3. Use 2 queries to avoid the subselect in the query which causes a lockup (#1242)

Note: I think you are missing the batch_id index on the errors.

### DIFF
--- a/Cron/BatchesClean.php
+++ b/Cron/BatchesClean.php
@@ -44,19 +44,6 @@ class BatchesClean
     }
     public function execute()
     {
-
-        try {
-            $connection = $this->mailChimpErrors->getResource()->getConnection();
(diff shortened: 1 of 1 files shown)


gonzaloebiz:
Hi @barryvdh 

thanks for your contribution.
Based in your contribution we make some modifications to your code that introduces some issues, like if the $existingBatchIds is an empty array (this cause an error)
You can look the modifications to your source in the commits associated to the issue.

Best

barryvdh:
Thank you!

Could you also look into the missing index for batch_id on the errors table?

PR #1296, opened 2021-09-07
closes #1288 for magento 2.3

### LINKED ISSUE #1288: Incorrect image url with some extensions (p.e. cloudinary extension)
Some image extensions don't fit the Magento way and return the image path without the preceding /, this cause that when we fill the product image to sync the URL is wrong

gonzaloebiz:


Juan-ebizmarts (approved):

cse00241:
Hi I can't unistall this extension can you please check it let me know why I can't unistall it??
I use install the extension with normal, app/code folder then when unistall but place order showing error.
Run compile setup:upgrade everything... 
There is some dependency let me know how I can unistall it.. I tried unistall:: not work..
Sale order place show error.

PR #1387, opened 2022-05-16
Declare optional parameters after required parameters.

claudio-ferraro:
Optional parameters should be declared after required parameters.


gonzaloebiz:
Please take a look at https://github.com/mailchimp/mc-magento2/issues/1377

PR #1440, opened 2022-06-26
Fix issue introduced by not selecting all quote fields

chrisastley:
Added extra fields to the select to fix the error generating on line 382 due to getCustomerEmail() and getUpdatedAt() being empty.


gonzaloebiz:
Hi @chrisastley 

thanks for your contribution, we are looking at this

gonzaloebiz:
Hi @chrisastley 

take a look at https://github.com/mailchimp/mc-magento2/issues/1460 (we already release a new version with this fix)

Best

PR #1459, opened 2022-07-11
Deprecated Functionality: abs(): Passing null to parameter

chrisastley:
Fix issue appearing in logs and preventing cron running when using PHP 8.1 Deprecated Functionality: abs(): Passing null to parameter #1 ($num) of type int|float is deprecated in vendor/mailchimp/mc-magento2/Model/Api/Order.php on line 374

### DIFF
--- a/Block/Adminhtml/System/Config/CreateAbandonedCart.php
+++ b/Block/Adminhtml/System/Config/CreateAbandonedCart.php
@@ -1,0 +1,38 @@
+<?php
+/**
+ * Ebizmarts_MailChimp Magento JS component
+ *
+ * @category    Ebizmarts
+ * @package     Ebizmarts_MailChimp
+ * @author      Ebizmarts Team <info@ebizmarts.com>
+ * @copyright   Ebizmarts (http://ebizmarts.com)
+ * @license     http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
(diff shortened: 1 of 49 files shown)


gonzaloebiz:
Hi @chrisastley,

thanks for your contributions.
First of all, I see that you make the change in magento 2.3, keep in mind that PHP 8.1 is not officially supported by magento 2.3 (take a look at [magento requirements](https://devdocs.magento.com/guides/v2.3/install-gde/system-requirements.html))
No matter this, your issue is not about casting the value to float, is about that you have nulls values for this field

Best

chrisastley:
We are on Magento 2.4.4 and there are no updates showing so I assumed this is the latest version. The float I added is for lines where no discount is returned on an order, are you saying it should return '0.00' instead of the null that is currently being returned?

gonzaloebiz:
Your pr is to the 2.3 branch, in any case must be to the 2.4 branch
Anyway, looking at the sales_order table, the discount_amount is always 0.00 if no discount is present, so there is no need to do this.
If you have other value than 0.00 on discount_amount for orders without discount, looks why

chrisastley:
Apologies for committing to the wrong branch, maybe you should change the default to 2.4 branch. I'll check the database and report back.

chrisastley:
Hi,

I've checked the database and there is a mixture of positive values, zero values and null. The field does allow in the database so I think my fix is relevant.

PR #1486, opened 2022-07-27
Remove unneccesary empty comment in db_schema.xml file

lumnn:
This module at the moment causes that `setup:db:status` returns `Declarative Schema is not up to date` message.

It's probably caused because the `db_schema.xml` is configured with empty comment for `updated_at` column. `setup:db:status` when comparing, sees that comment is `null` while `db_schema.xml` defines it should be an empty string.

This PR removes `comment` attribute entirely making `setup:db:status` work as expected.


gonzaloebiz (approved):

gonzaloebiz:
Thanks for your contribution @lumnn

PR #1599, opened 2022-11-07
Closes #1465 for 2.3

### LINKED ISSUE #1465: Create grid in admin that shows batch information
Basically, the idea is to create a grid that shows the following information about batches sent to Mailchimp:

-ID
-Mailchimp Store name
-Status
-#Carts
-#Customers
-#Orders
-#Products
-#Subscribers

New and modified items will be shown separately. It remains to be seen what exactly to do about items which aren't synced (with errors)

agustinbv:


gonzaloebiz (approved):

hostep:
@agustinbv, @gonzaloebiz: there is a typo here, either:
- the classname should be `GetResponse`
- the filename should be `Getresponse.php`

It's probably the second one you want since you use the url without a capital R: `mailchimp/batch/getresponse`

Error can be seen like this:
1. Install version 103.4.50 of mailchimp/mc-magento2
2. Run `composer dump-autoload -o`
3. You get this error: `Class Ebizmarts\MailChimp\Controller\Adminhtml\Batch\Getresponse located in ./vendor/mailchimp/mc-magento2/Controller/Adminhtml/Batch/GetResponse.php does not comply with psr-4 autoloading standard. Skipping.`

PR #1647, opened 2023-01-27
Added a float to abs() values, due to PHP 8.1 compatibility in cases …

numbology:
Added a float to abs() values, due to PHP 8.1 compatibility in cases where order discountAmount is null


gonzaloebiz:
Hi @numbology 

please add an issue with a complete description of the issue, you can find the issue reporting guidelines [here](https://github.com/mailchimp/mc-magento2/wiki/Issue-reporting-guidelines)

Best

Issue #1732, opened 2023-07-04
Cannot upgrade to latest version due to patch error


0xMatt:
## Issue Description

Getting this error when attempting to upgrade mailchimp:

```
Unable to apply data patch Ebizmarts\MailChimp\Setup\Patch\Data\Migrate452 for module Ebizmarts_MailChimp. Original exception message: Rolled back transaction has not been completed correctly.
```

### Preconditions

    1. Magento 2.4.6
    2. From Version: 103.4.51
    3. To Version: 103.4.52

### Steps to reproduce

    1. Run composer update while on version 103.4.51 which pulls in 103.4.52.
    2. Run bin/magento s:up

### Actual and Expected result

    Expected result:
    setup:upgrade succeeds without error

    Actual result:
    setup:upgrade fails with error in issue description

### Additional information

- PHP 8.1.20
- Mariadb 10.5.20

lytesaber:
Also getting the same data patch error, will have to roll back to the previous version until a fix has been provided.

`Unable to apply data patch Ebizmarts\MailChimp\Setup\Patch\Data\Migrate452 for module Ebizmarts_MailChimp. Original exception message: Rolled back transaction has not been completed correctly.`

PachisPachis:
Same here.

akosglue:
confirmed, m2 2.4.6, php 8.1

Garde85:
me too! This error appers to me on different magento installations. Magento 2.4.5-p3, php 8.1.

meetvora69:
me too! This error appears to me on different magento installations. Magento 2.4.6-p3, php 8.2

gonzaloebiz:
Hi 

we tested updating from 103.4.51 to 103.4.52 with no problems, even with a db-prefix and no problem at all. (we don't test it with split databases because was deprecated)
Perhaps if you replace the vendor/mailchimp/mc-magento2/Setup/Patch/Data/Migrate452.php with this one

```
<?php

namespace Ebizmarts\MailChimp\Setup\Patch\Data;

use Magento\Framework\Setup\ModuleDataSetupInterface;
use Magento\Framework\Setup\Patch\DataPatchInterface;
use Magento\Sales\Model\OrderFactory;
use Ebizmarts\MailChimp\Model\ResourceModel\MailChimpSyncEcommerce\CollectionFactory as SyncFactory;
use Ebizmarts\MailChimp\Helper\Data;

class Migrate452 implements DataPatchInterface
{
[...]

PR #1733, opened 2023-07-07
Fix issue #1732

0xMatt:
Fix for issue #1732


gonzaloebiz:
Hi @0xMatt i will take your solution with some improvement (i will add control to check if order exist or not), take a look at https://github.com/mailchimp/mc-magento2/commit/0e6c7782d02df970806152a36566efa98d9dbea1

Thanks for your contribution to fix it

PR #1856, opened 2023-11-07
Technical error when clicked on mailchimp column in customer grid

### LINKED ISSUE #1855: Something went wrong error appears when mailchimp column sorted in customer listing grid
## Issue Description
When we click on mailchimp column in the customer listing grid it throws an error stating “Something went wrong with processing the default view and we have restored the filter to its original state”. And the grid does not load.

image

### Preconditions

1. Module Version 103.4.57 (mailchimp/mc-magento2)
2. Magento Version 2.4.6-p2   

### Steps to reproduce

1. Clean shop with install sample data
2. Install “mailchimp/mc-magento2” module.
3. Go to Customers->All Customers.
4. Now click on Mailchimp column in the customer grid to sort it.

    Expected result:
    When the Mailchimp column in the customer grid is clicked, it should not throw an error.
    Actual result:
    When the Mailchimp column in the customer grid is clicked, it throws an error.

### Additional information

This issue occurs because when the Mailchimp column is clicked it goes into the database table ‘customer_grid_flat’ to search for the ‘mailchimp_async’ column which is not created and not present in the table.
Due to this, the error occurs.
It can be solved by disabling the sorting of the Mailchimp column in the customer grid.
Also, in other XML files such as “sales_order_grid.xml” and “product_listing.xml”
we can see that sortable false is given here.
But in “customer_listing.xml”, the sortable element is not present due to which it takes true by default. So using a sortable element in the customer listing file and giving it as false will disable sorting for the Mailchimp column in the customer listing grid.

### File name
mailchimp/mc-magento2/view/adminhtml/ui_component/customer_listing.xml

image

kapilpandya2203:
### Fixed Issues
Fixed: https://github.com/mailchimp/mc-magento2/issues/1855

### Description
When we click on mailchimp column in the customer listing grid it throws an error stating “Something went wrong with processing the default view and we have restored the filter to its original state”. And the grid does not load.

### Error screenshot
image

### Manual testing scenarios
1. Clean shop with install sample data
2. Install “mailchimp/mc-magento2” module.
3. Go to Customers->All Customers.
4. Now click on Mailchimp column in the customer grid to sort it.

### DIFF
--- a/view/adminhtml/ui_component/customer_listing.xml
+++ b/view/adminhtml/ui_component/customer_listing.xml
@@ -5,6 +5,7 @@
             <settings>
                 <label translate="true">Mailchimp</label>
                 <bodyTmpl>ui/grid/cells/html</bodyTmpl>
+                 <sortable>false</sortable>
             </settings>
         </column>
     </columns>


gonzaloebiz:
Look at #1981

PR #2289, opened 2026-06-05
Closes #2288 for Magento 2.4

### LINKED ISSUE #2288: Using a second Mailchimp account on a certain Magento scope makes the dropdown list disappear
If you have a Magento installation with the following scope structure:

-Website 1
  -Store View 1
-Website 2
  -Store View 2

And you have a certain Mailchimp account being used for Website 1/Store View 1, and you decide to set up another Mailchimp account on Website 2/Store View 2, the following happens:

-When you insert the API Key, the Mailchimp Stores and Lists dropdown menus load fine
-But then, selecting an specific Mailchimp Store generates an empty list dropdown, and an "Invalid API Key" error shown in "Account Details"
-Once you reload the page, the error message dissappears, but the dropdown list is still empty.
-The syncing works correctly, however, an API call error is reported at Store View 2 level: 
`Resource Not Found for Api Call: [https:\/\/us1.api.mailchimp.com\/3.0\/lists\/\/interest-categories?count=200] using method [GET
`
-Debug logs show an API call being made using the incorrect key at the website level when refreshing the page.

Comments:
- The API key and the Mailchimp Store selection is made at the Website level, while the Store View just uses the settings inherited from the Website level

agustinbv:
## Description

Fixes #2288.

When configuring Mailchimp on an additional **website scope** with an API Key belonging to a **different Mailchimp account**, selecting the Mailchimp Store showed an **"Invalid API Key"** message next to *Account Details* and left the **Mailchimp List dropdown empty**. The empty list id was then persisted, breaking the subscriber/interest sync for that store.

## Root cause

In `Controller/Adminhtml/Ecommerce/Getaccountdetails.php`, when the API Key field is masked (`******`), the front-end sends `encrypt = 3` together with the `scope`/`scopeId` being edited. The `encrypt == 3` branch resolved the API from the **admin store** instead of that scope:

    $api = $this->_helper->getApi($this->_storeManager->getStore()->getId());

In an admin request `getStore()` returns the admin store (id `0`), so the key is read at the **default scope** — never the website scope being configured. On the default scope that key is either the first account's key or empty (when Mailchimp is disabled at default), so the lookup against the second account's store fails with "Invalid API Key".

## Fix

Resolve the API from the scope passed by the front-end, consistent with the rest of the controller and the `MonkeyStore`/`MonkeyList` source models:

    $api = $this->_helper->getApi($scopeId, $scope);

## Affected file

- `Controller/Adminhtml/Ecommerce/Getaccountdetails.php`

## Testing

- Magento Open Source 2.4.5-p4, multi-website install with each website on a different Mailchimp account.
- In the second website scope (key already saved, field shows `******`), selecting the Mailchimp Store now populates Account Details and the Mailchimp List dropdown from the correct account; no "Invalid API Key".

### DIFF
--- a/Controller/Adminhtml/Ecommerce/Getaccountdetails.php
+++ b/Controller/Adminhtml/Ecommerce/Getaccountdetails.php
@@ -59,7 +59,7 @@ class Getaccountdetails extends Action
         $scopeId = $param['scopeId'];
         try {
             if ($encrypt == 3) {
-                $api = $this->_helper->getApi($this->_storeManager->getStore()->getId());
+                $api = $this->_helper->getApi($scopeId, $scope);
             } else {
                 $api = $this->_helper->getApiByApiKey($apiKey, $encrypt);
             }
--- a/Cron/Ecommerce.php
+++ b/Cron/Ecommerce.php
@@ -128,8 +128,8 @@ class Ecommerce
         foreach ($this->_storeManager->getStores() as $storeId => $val) {
             if ($this->_helper->getConfigValue(\Ebizmarts\MailChimp\Helper\Data::XML_PATH_ACTIVE, $storeId)) {
                 if (!$this->_ping($storeId)) {
-                    $this->_helper->log('MailChimp is not available');
-                    return;
+                    $this->_helper->log("MailChimp is not available for store $storeId, skipping");
+                    continue;
                 }
                 $this->_storeManager->setCurrentStore($storeId);
                 $listId = $this->_helper->getGeneralList($storeId);
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -646,7 +646,11 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
     public function getMCMinSyncDateFlag($storeId = null)
     {
         $syncDate = $this->getConfigValue(self::XML_PATH_IS_SYNC, $storeId);
-        if ($syncDate=='') {
+        // When a per-Mailchimp-store flag has been written under issync/<mailchimpStoreId>
+        // at default scope, this path resolves to an array for any store with no scalar
+        // override of its own. Such a store has not recorded a completion date, so treat
+        // it as "not yet synced" instead of returning the array to the callers.
+        if (is_array($syncDate) || $syncDate === null || $syncDate == '') {
             $syncDate = '1900-01-01';
         }
         return $syncDate;


agustinbv:
### Commit 7971f721 — Fix `getMCMinSyncDateFlag()` returning an array on multi-store installs

**Problem**

`saveMCMinSyncing()` overloads one config node, `mailchimp/general/issync`, two incompatible ways:

- with `$mailchimpStoreId = null` it writes a **scalar** at `mailchimp/general/issync` (per Magento store, the "fully-synced" date);
- with `$mailchimpStoreId` set it writes **children** at `mailchimp/general/issync/<mailchimpStoreId>` at default scope (per Mailchimp store, the "initial-sync" flag).

The moment one child exists, Magento treats `issync` as a **branch node**, so `_scopeConfig->getValue()` returns the whole subtree as an **array**. `getMCMinSyncDateFlag()` reads that parent path, so any active store that has not written its own scalar `issync` leaf (i.e. hasn't completed its initial sync) inherits the branch and gets an array instead of a date string. Callers consume it as a scalar (`generatePOSTPayload(... $isSynced)`, the `mailchimp_sync_delta < flag` comparisons, the cron gates), producing `Warning: Array to string conversion` and leaving non-default stores unable to build their payloads — they stay stuck "Syncing". A single child is enough to trigger it; the array's element count just reflects how many Mailchimp stores are configured.

**Fix**

Harden `getMCMinSyncDateFlag()` so it always returns a scalar: if the resolved value is an array (or null/empty), treat the store as "not yet synced" and return `1900-01-01`. That is the semantically correct answer — an array result means the store has no completion date of its own.

**Effect**

The store now builds its payloads normally, completes its initial sync, and writes its own scalar `issync` leaf, after which reads resolve to the scalar as usual — i.e. it self-heals. Handles already-poisoned data on live installs with no migration. Change is localized to the helper; `getMCMinSyncDateFlagByMailchimpStore()` (which reads a specific leaf) is untouched.

Verified locally on a 2-website / 2-account install: pre-fix the non-default store returned an array; post-fix it returns a scalar and a product, customer and order all synced to the second account with no errors.

agustinbv:
### Commit e0716956 — Skip unreachable store instead of aborting the whole ecommerce cron

**Problem**

In `Cron/Ecommerce::execute()`, the loop over `getStores()` did a hard `return` when a store's `_ping()` failed:

```php
if (!$this->_ping($storeId)) {
    $this->_helper->log('MailChimp is not available');
    return;
}
```

`return` exits the entire cron run, so a single store whose Mailchimp account is unreachable (bad/expired API key, transient API outage) aborts the loop and **every store after it is skipped for that run**. On a multi-store / multi-account install one bad account silently blocks all the others from ever syncing.

**Fix**

Use `continue` so the cron moves on to the next store, and log the store id being skipped:

```php
if (!$this->_ping($storeId)) {
    $this->_helper->log("MailChimp is not available for store $storeId, skipping");
    continue;
}
```

**Effect**

A failing account no longer stops the rest. Each store is now isolated for the purposes of the ping check, and the log message identifies which store was skipped instead of a generic line. This is independent of the `getMCMinSyncDateFlag()` fix in 7971f721 but is a real multi-store hazard, so it's bundled here.

gonzaloebiz (approved):

PR #2341, opened 2026-08-24
Report extension status to the ebizmarts service once an hour

gonzaloebiz:
Adds hourly status reporting to the ebizmarts service, so an installation can be seen as running without anyone opening the admin, and so messages can reach the merchant through Magento's own admin inbox.

## Why

The extension announces itself exactly once. `Model/Config/Source/Details.php` is a configuration source model, so it runs only when an administrator opens the Mailchimp configuration page, and it is gated on `if (!$token)` — it fires once per store view, ever, and then never speaks again.

That cannot tell an installation trading today from one abandoned years ago, it undercounts by construction because it needs a person to open a page, and it leaves no channel to reach the merchant.

## What this adds

An hourly cron branch: **register** once, then **report status** on every tick, and deliver any **notifications** the reply announces into the admin inbox.

| File | Role |
|---|---|
| `Model/Edge/Client` + `Response` | Performs the calls and classifies the outcome |
| `Model/Edge/LivenessSignals` | Is this store syncing, and is it failing — two indexed reads |
| `Model/Edge/Beacon` | Register once, report always, refresh account data twice a day |
| `Model/Edge/NotificationDelivery` | Writes into `adminnotification_inbox` via `NotifierInterface` |
| `Cron/EdgeBeacon` + `etc/crontab.xml` | Hourly entry point |

Everything is additive. The existing registration, the support-log push, the `mailchimp_notification` table and its cleanup, `enable_support` and the XML notification feed are untouched. No `system.xml` entry, no new table, no schema change.

## Decisions worth reviewing

**No customer data leaves the store, by construction.** `mailchimp_errors.errors` holds the raw error body and `mailchimp_notification.notification_data` holds request parameters and responses; both can carry customer details, and neither is ever selected. Only the surrounding facts travel — when, and what kind. A test asserts those columns are not in the select.

**Only a 401 discards the token.** A 429 and a 5xx both leave it untouched — the tick simply stops. `Retry-After` is read and logged but not obeyed: the next attempt is the next hourly tick whatever the header asked for, so nothing here schedules a back-off. If any other status discarded it, one bad deployment on the receiving side would turn into every installation re-registering at once. Tokens also expire, and that arrives as a 401, so expiry needs no separate handling.

**Null handling is asymmetric on purpose.** The receiving side deletes a field on an explicit null and keeps it when the key is absent. The error fields are therefore always present — null when there is no error row — so a recovered store actively clears its previous failure. `last_sync_at` is omitted when it cannot be read, so a transient local problem never erases a good value already held.

**A failing Mailchimp call never suppresses the report.** The account block rides along twice a day; an outage must not make every installation look dead at once.

**The cron entry carries both `<schedule>` and `<config_path>`.** The schedule is a bootstrap, not a duplicate. Magento skips a job whose expression resolves to nothing (`ProcessCronQueueObserver::_generateJobs`), and `config_path` is empty until the first successful registration writes it — which happens inside this very job. With `config_path` alone the job could never run and could never register. `17` rather than `0` so a not-yet-registered install does not land on the minute the rest of Magento's cron already piles up on.

**The minute is written in default scope.** The scheduler reads `config_path` with no store id, so a value written per store view would never be seen and the job would sit on its bootstrap minute forever, looking configured. The first store view to register claims it; later ones leave it alone.

**Notifications degrade rather than fatal.** `NotifierInterface` resolves through `app/etc/di.xml` to the framework's `NotifierPool`, which iterates a list that is empty when `Magento_AdminNotification` is disabled, so every call becomes a no-op.

**Delivery is confirmed.** The ids that reached the inbox are echoed on the next report and cleared only once a report has actually returned 200 — one that never landed has confirmed nothing. The queue appends rather than overwrites, so a batch still awaiting acknowledgement is not dropped by a newer one. Nothing claims the merchant opened anything; only that the message reached the inbox.

## Tests

35 unit tests, 55 assertions: the status matrix (including that a 5xx and a transport failure are **not** treated as auth failures), the account-refresh cadence and that a failing account call still reports, the acknowledgement queue surviving a failed report, the priority mapping, and that no customer-bearing column is selected.

Verified end to end against a live install with the cron running unattended, not only in unit tests.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,9 @@
+# Keep development-only paths out of release archives.
+#
+# `export-ignore` applies to `git archive`, which is what GitHub builds release
+# tarballs from and what Composer downloads on a dist install. It does NOT
+# affect `git clone`, so a developer cloning the repo — or an app/code install
+# done by cloning — still gets everything listed here.
+
+/Test               export-ignore
+/.gitattributes     export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,9 +1,0 @@
-# Keep development-only paths out of release archives.
-#
-# `export-ignore` applies to `git archive`, which is what GitHub builds release
-# tarballs from and what Composer downloads on a dist install. It does NOT
-# affect `git clone`, so a developer cloning the repo — or an app/code install
-# done by cloning — still gets everything listed here.
-
-/Test               export-ignore
-/.gitattributes     export-ignore
--- a/Cron/EdgeBeacon.php
+++ b/Cron/EdgeBeacon.php
@@ -1,0 +1,42 @@
+<?php
+/**
+ * Ebizmarts_MailChimp Magento Component
+ *
+ * @category    Ebizmarts
+ * @package     Ebizmarts_MailChimp
+ * @author      Ebizmarts Team <info@ebizmarts.com>
+ * @copyright   Ebizmarts (http://ebizmarts.com)
+ * @license     http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+namespace Ebizmarts\MailChimp\Cron;
+
+use Ebizmarts\MailChimp\Model\Edge\Beacon;
+
+/**
+ * Hourly entry point for the status reporting beacon.
+ */
+class EdgeBeacon
+{
+    /**
+     * @var Beacon
+     */
+    private $beacon;
+
+    /**
+     * @param Beacon $beacon
+     */
+    public function __construct(
+        Beacon $beacon
+    ) {
+        $this->beacon = $beacon;
+    }
+
+    /**
+     * @return void
+     */
+    public function execute()
+    {
+        $this->beacon->execute();
+    }
+}
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -539,6 +539,46 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
     }
 
     /**
+     * Write a config value now and leave the cache flush to the caller.
+     *
+     * saveConfigValue() flushes on every call, and a flush is not cheap: the
+     * config module intercepts it to re-read the whole system config and then
+     * serialise and encrypt every scope in turn, under a lock. That cost grows
+     * with the number of store views, so a job that writes once per view pays
+     * it once per view — quadratic work for what should be one refresh.
+     *
+     * Callers that write several values in one pass should use this and then
+     * call flushConfigCache() once. Nothing they write is visible to a read
+     * until they do, so this is only safe where the caller does not read back
+     * what it has just written.
+     *
+     * @param  string      $path
+     * @param  string      $value
+     * @param  int|null    $storeId
+     * @param  string|null $scope
+     * @return void
+     */
+    public function saveConfigValueWithoutCacheFlush($path, $value, $storeId = null, $scope = null)
+    {
+        $this->saveConfigValueAtomic(
+            (string)$path,
+            (string)$value,
+            (string)($scope ?: \Magento\Store\Model\ScopeInterface::SCOPE_STORES),
+            (int)$storeId
+        );
+    }
+
+    /**
+     * Refresh the config cache once, after a run of deferred writes.
+     *
+     * @return void
+     */
+    public function flushConfigCache()
+    {
+        $this->_cacheTypeList->cleanType('config');
+    }
+
+    /**
      * Atomically upsert a single config row using INSERT … ON DUPLICATE KEY UPDATE.
      *
      * core_config_data has a UNIQUE KEY on (scope, scope_id, path), so a plain
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -73,9 +73,17 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
 
     /**
      * Whether the merchant lets the account owner's name and address travel
-     * with diagnostics. Shared with the API library, which reads the same path
-     * for the same purpose: one switch in the admin has to govern both, or it
-     * tells the merchant something that is not true.
+     * with diagnostics.
+     *
+     * The path is shared with the API library on purpose: one switch in the
+     * admin has to govern every lane that reports, or it describes only part
+     * of what leaves the server.
+     *
+     * The library reads it from the release that first gives it anything to
+     * report — the constant and the reporting arrived in the same commit
+     * there, so no version of it can send contact details while ignoring this.
+     * Until that release is the one installed, the switch governs this
+     * extension's reports and there is nothing else for it to govern.
      */
     const XML_TELEMETRY_SHARE_CONTACT = 'mailchimp/telemetry/share_contact';
 
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -61,6 +61,16 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
     const XML_REGISTER_URL            = 'mailchimp/statistics/register_url';
     const XML_STATISTICS_TOKEN        = 'mailchimp/register/token';
 
+    /**
+     * Status reporting beacon. The token is per store view; the cron expression is
+     * per install; the delivery uid is the last notification actually written
+     * to the admin inbox. Deliberately NOT mailchimp/register/token, which
+     * holds a token from the previous relay and is meaningless to this service.
+     */
+    const XML_EDGE_TOKEN              = 'mailchimp/register/edge_token';
+    const XML_EDGE_BEACON_CRON        = 'mailchimp/register/beacon_cron';
+    const XML_EDGE_DELIVERY_UID       = 'mailchimp/register/last_delivery_uid';
+
     const XML_PIXEL_ENABLED_FOR_STORE = 'mailchimp/pixel/enabled_for_store';
     const XML_PIXEL_SCRIPT_URL        = 'mailchimp/pixel/script_url';
     const XML_PIXEL_SCRIPT_FRAGMENT   = 'mailchimp/pixel/script_fragment';
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -71,6 +71,14 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
     const XML_EDGE_BEACON_CRON        = 'mailchimp/register/beacon_cron';
     const XML_EDGE_DELIVERY_UID       = 'mailchimp/register/last_delivery_uid';
 
+    /**
+     * Whether the merchant lets the account owner's name and address travel
+     * with diagnostics. Shared with the API library, which reads the same path
+     * for the same purpose: one switch in the admin has to govern both, or it
+     * tells the merchant something that is not true.
+     */
+    const XML_TELEMETRY_SHARE_CONTACT = 'mailchimp/telemetry/share_contact';
+
     const XML_PIXEL_ENABLED_FOR_STORE = 'mailchimp/pixel/enabled_for_store';
     const XML_PIXEL_SCRIPT_URL        = 'mailchimp/pixel/script_url';
     const XML_PIXEL_SCRIPT_FRAGMENT   = 'mailchimp/pixel/script_fragment';
--- a/Model/Edge/Beacon.php
+++ b/Model/Edge/Beacon.php
@@ -218,7 +218,11 @@ class Beacon
             // Mailchimp outage must not make every installation look dead.
             $account = $this->accountBlock($storeId);
             if ($account !== null) {
-                $body['account'] = $account;
+                // Top level, not nested: the receiving side reads these as
+                // first-class keys and drops anything it does not recognise, so
+                // an `account` wrapper would be accepted with a 200 and thrown
+                // away. Same shape the register already sends.
+                $body = array_merge($body, $account);
             }
         }
 
@@ -318,7 +322,11 @@ class Beacon
      */
     private function shouldSendAccountBlock($storeId, $storeUrl)
     {
-        $slot = crc32($storeId . $storeUrl) % self::ACCOUNT_BLOCK_PERIOD_HOURS;
+        // Sign-masked: crc32() returns a NEGATIVE int on 32-bit PHP builds, and a
+        // negative slot can never equal an hour, so those installs would never
+        // refresh their account data. Silent, and 32-bit builds still exist on
+        // older shared hosting.
+        $slot = (crc32($storeId . $storeUrl) & 0x7fffffff) % self::ACCOUNT_BLOCK_PERIOD_HOURS;
         $hour = (int)$this->helper->getGmtDate('G');
 
         return $hour % self::ACCOUNT_BLOCK_PERIOD_HOURS === $slot;
@@ -349,7 +357,9 @@ class Beacon
 
         // Concatenation, not addition: on PHP 8 "abc" + "def" is a TypeError,
         // and for all-numeric ids it would silently produce a different slot.
-        $minute = crc32($accountId . $storeUrl) % 60;
+        // Sign-masked for the same reason as the slot above: a negative minute
+        // writes an invalid cron expression, and the job then never runs at all.
+        $minute = (crc32($accountId . $storeUrl) & 0x7fffffff) % 60;
 
         $this->helper->saveConfigValue(
             MailChimpHelper::XML_EDGE_BEACON_CRON,
--- a/Model/Edge/Beacon.php
+++ b/Model/Edge/Beacon.php
@@ -68,6 +68,11 @@ class Beacon
     private $scopeConfig;
 
     /**
+     * @var bool the cron slot has been settled for this process
+     */
+    private $slotClaimed = false;
+
+    /**
      * @param StoreManagerInterface    $storeManager
      * @param MailChimpHelper          $helper
      * @param Client                   $client
@@ -256,7 +261,10 @@ class Beacon
             );
         }
 
-        $this->notifications->handle($storeId, $token, $response->getData());
+        // The acks just sent are gone as far as the service is concerned, so
+        // nothing is still pending. Passed in rather than re-read: see
+        // NotificationDelivery::rememberForAck.
+        $this->notifications->handle($storeId, $token, $response->getData(), []);
     }
 
     /**
@@ -350,8 +358,21 @@ class Beacon
      */
     private function claimBeaconSlot($accountId, $storeUrl)
     {
+        // The read below cannot be trusted to see a write made earlier in this
+        // same process: within one request the config a scope has already
+        // loaded is refreshed only as a side effect of the cache-warming path,
+        // which is skipped when the config cache type is disabled. On a fresh
+        // multi-store-view install every view registers on the first tick, each
+        // read returns empty, and each writes a different minute — the last one
+        // winning, after N full config-cache flushes. This flag makes the claim
+        // once per process regardless.
+        if ($this->slotClaimed) {
+            return;
+        }
+
         $existing = $this->scopeConfig->getValue(MailChimpHelper::XML_EDGE_BEACON_CRON);
         if ($existing) {
+            $this->slotClaimed = true;
             return;
         }
 
@@ -367,6 +388,8 @@ class Beacon
             0,
             'default'
         );
+
+        $this->slotClaimed = true;
     }
 
     /**
--- a/Model/Edge/Beacon.php
+++ b/Model/Edge/Beacon.php
@@ -268,6 +268,31 @@ class Beacon
     }
 
     /**
+     * Whether the account owner's name and address may travel with a report.
+     *
+     * Only an explicit no counts as a refusal. An absent value means the
+     * question has never been answered on this installation — which is what an
+     * upgrade from a version that predates the switch looks like — and that is
+     * not the same as declining. config.xml supplies the default, so in
+     * practice the absent case only arises if configuration cannot be read at
+     * all, and failing open there matches how the API library reads the very
+     * same path.
+     *
+     * @param  int $storeId
+     * @return bool
+     */
+    private function contactAllowed($storeId)
+    {
+        $value = $this->helper->getConfigValue(MailChimpHelper::XML_TELEMETRY_SHARE_CONTACT, $storeId);
+
+        if ($value === null || $value === '') {
+            return true;
+        }
+
+        return (bool)$value;
+    }
+
+    /**
      * Delivery uids waiting to be acknowledged, queued by the last pull.
      *
      * @param  int $storeId
@@ -303,13 +328,18 @@ class Beacon
             return null;
         }
 
+        $fields = ['account_id', 'account_name', 'pricing_plan_type', 'total_subscribers'];
+
+        // Everything that names or reaches the person behind the account is
+        // sent only while the merchant allows it. The admin switch says it
+        // leaves the owner's name and address out of these reports, and a
+        // switch that is not obeyed is worse than no switch at all.
+        if ($this->contactAllowed($storeId)) {
+            $fields = array_merge($fields, ['email', 'first_name', 'last_name', 'username']);
+        }
+
         $block = [];
-        foreach (
-            [
-                'account_id', 'account_name', 'email', 'first_name',
-                'last_name', 'username', 'pricing_plan_type', 'total_subscribers',
-            ] as $field
-        ) {
+        foreach ($fields as $field) {
             if (isset($root[$field])) {
                 $block[$field] = $root[$field];
             }
--- a/Model/Edge/Beacon.php
+++ b/Model/Edge/Beacon.php
@@ -1,0 +1,386 @@
+<?php
+/**
+ * Ebizmarts_MailChimp Magento Component
+ *
+ * @category    Ebizmarts
+ * @package     Ebizmarts_MailChimp
+ * @author      Ebizmarts Team <info@ebizmarts.com>
+ * @copyright   Ebizmarts (http://ebizmarts.com)
+ * @license     http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+namespace Ebizmarts\MailChimp\Model\Edge;
+
+use Ebizmarts\MailChimp\Helper\Data as MailChimpHelper;
+use Magento\Framework\App\Config\ScopeConfigInterface;
+use Magento\Framework\App\ProductMetadataInterface;
+use Magento\Store\Model\ScopeInterface;
+use Magento\Store\Model\StoreManagerInterface;
+
+/**
+ * Hourly status report to the ebizmarts service: register once, then ping forever.
+ *
+ * Additive by design. Nothing here touches the existing registration to
+ * users-mc4mage, the hourly batch push or enable_support.
+ */
+class Beacon
+{
+    /**
+     * The account block rides two pings a day, twelve hours apart. The slot is
+     * derived from the store identity rather than persisted, so there is no
+     * "last refresh" timestamp to write on every tick.
+     */
+    const ACCOUNT_BLOCK_PERIOD_HOURS = 12;
+
+    /**
+     * @var StoreManagerInterface
+     */
+    private $storeManager;
+
+    /**
+     * @var MailChimpHelper
+     */
+    private $helper;
+
+    /**
+     * @var Client
+     */
+    private $client;
+
+    /**
+     * @var LivenessSignals
+     */
+    private $signals;
+
+    /**
+     * @var NotificationDelivery
+     */
+    private $notifications;
+
+    /**
+     * @var ProductMetadataInterface
+     */
+    private $productMetadata;
+
+    /**
+     * @var ScopeConfigInterface
+     */
+    private $scopeConfig;
+
+    /**
+     * @param StoreManagerInterface    $storeManager
+     * @param MailChimpHelper          $helper
+     * @param Client                   $client
+     * @param LivenessSignals          $signals
+     * @param NotificationDelivery     $notifications
+     * @param ProductMetadataInterface $productMetadata
+     * @param ScopeConfigInterface     $scopeConfig
+     */
+    public function __construct(
+        StoreManagerInterface $storeManager,
+        MailChimpHelper $helper,
+        Client $client,
+        LivenessSignals $signals,
+        NotificationDelivery $notifications,
+        ProductMetadataInterface $productMetadata,
+        ScopeConfigInterface $scopeConfig
+    ) {
+        $this->storeManager    = $storeManager;
+        $this->helper          = $helper;
+        $this->client          = $client;
+        $this->signals         = $signals;
+        $this->notifications   = $notifications;
+        $this->productMetadata = $productMetadata;
+        $this->scopeConfig     = $scopeConfig;
+    }
+
+    /**
+     * One tick: every store view that has an API key and an active extension.
+     *
+     * A failure on one store view never aborts the loop.
+     *
+     * @return void
+     */
+    public function execute()
+    {
+        foreach ($this->storeManager->getStores() as $store) {
+            $storeId = (int)$store->getId();
+
+            if (!$this->helper->isMailChimpEnabled($storeId) || !$this->helper->getApiKey($storeId)) {
+                continue;
+            }
+
+            try {
+                $this->processStore($storeId, (string)$store->getBaseUrl());
+            } catch (\Exception $e) {
+                $this->helper->log('Edge beacon failed for store ' . $storeId . ': ' . $e->getMessage());
+            }
+        }
+    }
+
+    /**
+     * Register when there is no token, then ping regardless.
+     *
+     * @param  int    $storeId
+     * @param  string $storeUrl
+     * @return void
+     */
+    private function processStore($storeId, $storeUrl)
+    {
+        $token = (string)$this->helper->getConfigValue(MailChimpHelper::XML_EDGE_TOKEN, $storeId);
+
+        if ($token === '') {
+            $token = $this->register($storeId, $storeUrl);
+            if ($token === null) {
+                // Nothing written. The remote upsert is idempotent, so the next
+                // hourly tick simply tries again.
+                return;
+            }
+        }
+
+        $this->ping($storeId, $storeUrl, $token);
+    }
+
+    /**
+     * Register the store view and store the token it returns.
+     *
+     * The account fields cost one GET / against Mailchimp, and only here.
+     *
+     * @param  int    $storeId
+     * @param  string $storeUrl
+     * @return string|null
+     */
+    private function register($storeId, $storeUrl)
+    {
+        $account = $this->accountBlock($storeId);
+        if ($account === null) {
+            // Without the account we have no identity to register under.
+            return null;
+        }
+
+        $body = array_merge(
+            ['store_url' => $storeUrl],
+            $account,
+            $this->versionBlock()
+        );
+
+        $response = $this->client->register($body);
+        if (!$response->isOk()) {
+            return null;
+        }
+
+        $data  = $response->getData();
+        $token = isset($data['token']) ? (string)$data['token'] : '';
+        if ($token === '') {
+            $this->helper->log('Edge beacon registered store ' . $storeId . ' but no token came back');
+            return null;
+        }
+
+        $this->helper->saveConfigValue(
+            MailChimpHelper::XML_EDGE_TOKEN,
+            $token,
+            $storeId,
+            ScopeInterface::SCOPE_STORES
+        );
+
+        $this->claimBeaconSlot(isset($account['account_id']) ? (string)$account['account_id'] : '', $storeUrl);
+
+        return $token;
+    }
+
+    /**
+     * Send the status report and act on the reply.
+     *
+     * @param  int    $storeId
+     * @param  string $storeUrl
+     * @param  string $token
+     * @return void
+     */
+    private function ping($storeId, $storeUrl, $token)
+    {
+        $body = array_merge(
+            ['store_url' => $storeUrl],
+            $this->versionBlock(),
+            $this->signals->forStore($storeId)
+        );
+
+        // Two-step ack, opportunistic: the service acknowledges each uid
+        // included in the ping. `read[]` is deliberately
+        // never sent - we can prove a message reached the inbox, not that the
+        // merchant opened it.
+        $pendingAcks = $this->pendingAcks($storeId);
+        if (!empty($pendingAcks)) {
+            $body['received'] = $pendingAcks;
+        }
+
+        if ($this->shouldSendAccountBlock($storeId, $storeUrl)) {
+            // Deliberately outside any try that wraps the ping itself: a
+            // Mailchimp outage must not make every installation look dead.
+            $account = $this->accountBlock($storeId);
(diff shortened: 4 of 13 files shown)


dipola:
Two things before this merges, both the same species as the `extension_version` trap the Beacon docblock already documents:

1. **The account block is dropped on arrival.** `Beacon.php` nests it (`$body['account'] = $account`), but the service reads those fields as top-level keys and discards anything it does not recognise. The ping still returns 200 and the account data silently goes nowhere. Merge the account fields into the top level of the body (and the tests asserting `$body['account'][...]` flip to top-level too).

2. **`crc32()` needs the sign mask at both `% N` sites.** On 32-bit PHP builds `crc32()` returns negatives, so in `shouldSendAccountBlock()` a negative slot never matches the hour, and in the cron-minute site a negative minute writes an invalid cron expression, so those installs never report at all. Both failures are silent. `(crc32($x) & 0x7fffffff) % N` fixes both, and it is worth doing because older shared hosting still runs 32-bit builds.

gonzaloebiz:
Both fixed in `7ace3042`. You were right on both, and the first one has an observable fingerprint I had already seen and misread.

## The account block

Fixed — merged into the top level of the body, same shape the registration already used.

The tell was sitting in the data. The registration sends those fields flat (`array_merge(['store_url' => …], $account, …)`); only the report nested them. So the account values landed once, on the day this store first registered, and then never moved again — while `updated_at` kept advancing on every report. I saw exactly that on a live row and attributed all of it to a different cause. The nesting was the reason the twice-daily refresh had been posting into nothing.

Tests that asserted `$body['account'][…]` now assert the flat keys, and also assert that no `account` wrapper is present, so a regression fails loudly instead of going quiet.

## The sign mask

Fixed at both sites, and it earned a test: the claimed cron expression is now pinned to `/^([0-9]|[1-5][0-9]) \* \* \* \*$/`, which is what would have caught a negative minute.

**One consequence worth stating, because it is not only a 32-bit fix.** On 64-bit builds `crc32()` already returns a positive int, but masking clears the high bit, so it changes the derived value whenever that bit is set — roughly half of inputs. Measured on this store:

```
crc32 = 2239474935 (0x857BACF7), high bit set
unmasked % 12 = 3      masked % 12 = 7
```

So the hour slot re-shuffles for about half the estate the moment this ships. Harmless in itself — the slot is arbitrary and the point is only that installs do not land together — but it is a behaviour change on 64-bit, not a no-op that only rescues 32-bit hosts.

The persisted cron minute does **not** re-shuffle, which is the half that would have mattered: `claimBeaconSlot()` returns early when a value already exists, so an install that has registered keeps the minute it was given. Only new registrations derive under the new formula. No fleet-wide cron movement.

36 unit tests green.

agustinbv:
## Review of `7ace3042`

Reviewed against a fresh clone at `7ace3042`, diffed against `git merge-base HEAD origin/develop-2.4` (= `fbda9095`). Flagging up front that the branch was force-pushed mid-review — an earlier pass against `f772bfb8` reached different conclusions, so the ref matters here.

The design is solid and the reasoning in the description is unusually good — the `<schedule>` + `<config_path>` bootstrap, the default-scope write for the cron minute, and the asymmetric null handling are all correct and I verified each against the framework. The notes below are the exceptions.

---

### Blocking: the branch deletes `.gitattributes`

```
git diff $(git merge-base HEAD origin/develop-2.4) HEAD -- .gitattributes
```

`d96c7190` removes the file entirely, reverting #2342. `/Test export-ignore` goes with it, so `Test/` ships in release tarballs and Composer dist installs again. Looks like a rebase artifact from the force-push rather than anything intended — nothing in this PR's scope touches packaging.

Related: the description says "Everything is additive" and the pre-force-push stats said 0 deletions. The current diff is 13 files, 1948 insertions, **9 deletions**. Worth refreshing the description once this is restored.

---

### Confirmed defects

**1. The ack queue never actually empties.**

`Beacon::ping()` clears `XML_EDGE_DELIVERY_UID`, then calls `NotificationDelivery::handle()` → `rememberForAck()`, which reads that same path back. That read returns the pre-clear value.

`Helper::saveConfigValue` ends in `cleanType('config')`, which routes through `Magento\Config\Plugin\Framework\App\Cache\TypeList\WarmConfigCache::aroundCleanType` into `Magento\Config\App\Config\Type\System::cleanAndWarmDefaultScopeData()` — and **neither branch of that method resets `$this->data`**. It refills the cache backend; the in-process array survives. On the read side, `System::getWithParts()` guards with `if (!isset($this->data[$scopeType][$scopeId]))`, so a scope already loaded this request is never re-read.

Net: on any tick that both acks an old batch and delivers a new one, the old uids are merged back in and re-sent as `received[]` indefinitely (capped at 50 by `MAX_PENDING_ACKS`). The ordering comment — "Cleared only once the service has actually seen them, and before the pull below can queue a fresh batch" — describes the right intent, but the write doesn't take effect in time for the read.

Suggested fix: pass the cleared list into `handle()` rather than having `rememberForAck()` re-read config.

**Why the tests don't catch it.** `BeaconTest` mocks `NotificationDelivery` wholesale, so the clear-then-read seam is never exercised end to end. `NotificationDeliveryTest` covers `rememberForAck` in isolation behind `$helper->method('getConfigValue')->willReturn('uid-0,uid-1')` — a static stub with no write-then-read coupling. `testDeliveredUidsAppendToWhatIsStillPending` therefore asserts `'uid-0,uid-1,uid-2'`, which is exactly the stale-read outcome. The test encodes the bug as correct behaviour. Worth one test that drives `Beacon::ping()` with a real `NotificationDelivery` and a helper double that models config as write-then-stale-read.

**2. An empty-subject notification is acked without being delivered.**

In `NotificationDelivery::handle`, `deliver()` returns silently when `$title === ''`, but control falls through to `$delivered[] = $item['id']`. Only an exception skips the ack. That contradicts the stated contract ("we can prove a message reached the inbox").

Verified by a throwaway test: with `subject: ""`, `NotifierInterface` is never called and `uid-empty` is still queued for ack. One-line fix — have `deliver()` return a bool and gate the ack on it.

**3. `claimBeaconSlot` writes once per store view, not once per install.**

Same root cause as (1). On a fresh install with N store views, all N register in the first tick; each one's `scopeConfig->getValue()` returns stale-empty, so each writes a *different* minute and the last wins. Self-correcting from tick 2 onward, so impact is low — but "one install, one slot, one write … the value never flaps" isn't what happens on the first tick, and it's N full config-cache flushes instead of one.

---

### Worth a look — reasoned from code, not executed

**4. Config cache flush storm on token expiry.** `saveConfigValue` flushes the *entire install's* config cache, and `WarmConfigCache` then re-reads and re-caches the whole system config under a lock. Steady state writes nothing, so the common path is clean. But the description notes tokens expire and the service signals that as `401` — when a wave expires, that's a clear-flush plus a re-register-flush per store view per hour. On a large multi-store-view install that's a lot of lock contention on the normal expiry path, not an edge case.

**5. `getLastError` has no index supporting its sort.** `WHERE store_id = ? ORDER BY id DESC LIMIT 1` against `mailchimp_errors`, whose only index is `(store_id, regtype, original_id)`. InnoDB appends the PK, so rows for a store_id are ordered by `(regtype, original_id, id)` — not `id`. MySQL filesorts, or scans the PK backwards. I also couldn't find any pruning for that table (only `Setup/Uninstall`), so on a store with chronic sync errors it grows unbounded and this runs hourly per store view. "Holds few rows" is an assumption rather than a guarantee.

**6. Duplicate admin-inbox entries across store views.** `adminnotification_inbox` is global, but the beacon pulls per store view with a per-store-view token. If the service broadcasts one message to every registered store view of an install, the merchant sees N copies in the bell. Entirely dependent on how the service targets messages — flagging it as a question for the platform side rather than a defect here.

**7. `Retry-After` is parsed and stored but never consumed.** `Response::getRetryAfter()` has no callers; the 429 branch just returns. Since the next attempt is an hour out that's mostly fine, but "a 429 backs off" oversells it, and a `Retry-After` longer than an hour is ignored.

---

### Verified as correct

Checked against the installed 2.4 core and the module source, all good:

- `crontab.xsd` declares `schedule` then `config_path` as an `xs:sequence` — the ordering note is right, and reversing it would be a hard validation failure.
- `config_path` beats `schedule` only when non-empty, so the bootstrap-minute reasoning holds.
- `NotifierInterface` → `NotifierPool` via `app/etc/di.xml`, and `addCritical/addMajor/addNotice($title, $description, $url = '', ...)` match the 3-arg calls. The degrade-to-no-op claim is accurate.
- `Model/HTTP/Client/Curl` has every method `Client` uses, `makeRequest` genuinely throws on transport failure (so the catch isn't dead code), `setHeaders` takes the assoc array shape passed, and `parseHeaders` keys by header name so the `strtolower` scan in `readRetryAfter` works.
- Token written with `ScopeInterface::SCOPE_STORES` (plural) — correct, and notably avoids the singular-`'website'` scope bug that exists in the legacy `Details.php` path.
- All schema columns exist: `mailchimp_sync_ecommerce.mailchimp_sync_delta`, `mailchimp_errors.{id,type,status,added_at,store_id}`.
- crc32 sign-masking at both sites is right, and the top-level account fields match what register already sends.

**Test run:** 36 tests, 57 assertions, 0 failures on PHP 8.4. (`Ebizmarts\...\CurlFactory` needed a stub to run outside a generated Magento install — harness gap, not a defect.)

gonzaloebiz:
Thanks — this was a genuinely useful review, and the `.gitattributes` catch in particular is one I should have made myself. All four items are addressed in `981c0304`, `71c1508c` and `51946c3c`. **Pushed as new commits, not a force-push**, so `7ace3042` is still an ancestor and your review point does not move again.

## Blocking: `.gitattributes` — restored

Right on every count, including that nothing in this PR's scope touches packaging. Restored byte for byte from `develop-2.4`. The diff against the merge base is back to **12 files, 2230 insertions, 0 deletions**, so "everything is additive" holds again without editing the description.

The cause is in the reflog, and it was not the rebase:

```
fbda9095 @{4}: reset: moving to origin/develop-2.4
d96c7190 @{3}: commit: Report extension status to the ebizmarts service once an hour
```

The history was squashed by resetting onto `develop-2.4` and recommitting. `.gitattributes` had landed on `develop-2.4` *after* this branch was cut, so the working tree never had it; the reset moved HEAD onto a commit that did, and staging everything recorded the difference as a deliberate deletion. Same staleness produced a second symptom, below.

## Defect 2 — empty subject: confirmed, fixed

No argument. `deliver()` returned `void` and control fell through to the ack. It now returns whether it wrote, and the ack is gated on that. Covered by two tests: one that an empty-subject item is neither delivered nor confirmed, and one that it does not stop the rest of the batch.

## Defects 1 and 3 — real, but one step of the mechanism is off

You traced the right classes, and the fix you suggested is the right fix — I took it. But the conclusion that the ack queue "never actually empties" is not what happens on a default install, and the reason is one line inside the warm path.

`cleanAndWarmDefaultScopeData()`'s collector calls `readData()`, and `readData()` opens with:

```php
$this->data = $this->reader->read();
```

So it *does* refresh the in-process array — not by an explicit reset, but as a side effect of reading. Measured inside a running 2.4.8 by writing a unique value and reading it back in the same process:

| config cache | read after write | |
|---|---|---|
| **enabled** (the default) | `'uid-41910-6720'` | fresh — the defect does not manifest |
| **disabled** | `''` instead of the seed | **stale — the defect is real** |

With the cache disabled, `cleanAndWarmDefaultScopeData()` takes its `$cleaner(); return;` branch, never calls `readData()`, and the scope keeps what it was first loaded with. That is exactly your scenario; it is just gated on a configuration rather than universal.

Both are fixed anyway, because depending on how an install has its cache configured is not a property worth having:

- **Ack queue:** what is still pending is now an argument to `handle()`, not a config read-back. Same behaviour either way.
- **Cron slot:** a process-level flag claims it once, so N store views on a first tick write one minute instead of N.

**Your point 4 comes out stronger, not weaker.** Since every `saveConfigValue` reaches `cleanAndWarmDefaultScopeData()`, each write re-reads the entire config, then serializes *and encrypts* every scope before saving — under the lock. The steady-state case is still clean because nothing is written, but the token-expiry wave is more expensive than "a flush".

## On the tests

You were right that `testDeliveredUidsAppendToWhatIsStillPending` encoded the bug as correct behaviour. Corrected, and five added — including the one you asked for: two drive `Beacon::ping()` through a **real** `NotificationDelivery` with a helper double that models config as write-then-stale-read. Mocking the delivery wholesale is precisely why the seam was never exercised.

All five fail against the code before this change — verified by reverting the two model files and re-running: 6 failures, then green. **41 edge tests, 64 assertions.**

I also merged `develop-2.4` in. The four `PixelStateWriterTest` failures you would have hit running the whole module suite were the same staleness as the `.gitattributes` deletion — the corrections had landed on the base branch after this one was cut. Full module suite is now **116 tests, 213 assertions, green**.

## Points 5, 6 and 7

Not addressed here, deliberately:

- **5 (`getLastError` sort)** — agreed, the index does not support it and "holds few rows" is an assumption. Worth its own change rather than being buried in this one.
- **6 (duplicate inbox entries)** — agreed it is a question for the receiving side, not a defect here. Raising it there.
- **7 (`Retry-After` unused)** — correct, `getRetryAfter()` has no callers. I would rather fix the claim than add machinery that runs once an hour anyway.

The `CurlFactory` stub you needed is a real harness gap and worth fixing separately.

gonzaloebiz:
## The contact opt-out has to be honoured here too

Raising this myself after #2344 merged, because the two together would have shipped a switch that is false the moment a merchant uses it.

The Diagnostics field tells the merchant:

> Set this to **No** to leave the account owner's name and email address out of those reports; everything else keeps working.

The API library obeys it — it reads `mailchimp/telemetry/share_contact` and gates the owner fields on the answer. **This report did not.** `accountBlock()` collected `email`, `first_name`, `last_name` and `username` unconditionally, and sent them on registration and on every account refresh afterwards, an hour apart. Nothing in this branch consulted the switch.

That is worse than not offering a switch, because the merchant acts on it.

Fixed in `d1186082`. The four fields are sent only while the switch allows it; `account_id`, `account_name`, the plan and the subscriber count are unaffected, exactly as the wording promises. The config path is shared with the library on purpose — one control in the admin has to govern both, or it describes only half of what leaves the server.

An unanswered switch is read as permission rather than refusal: an installation upgrading from a version that predates the field has never been asked, and that is not the same as declining. `config.xml` supplies the default, so this only arises when configuration cannot be read at all, and the library reads the same path the same way.

Three tests — declining leaves all four out while the rest still travels, allowing sends them, and an unanswered switch is not a refusal. The first fails against the code before the change. **119 tests, 226 assertions, green.**

### One inconsistency worth a decision, not fixed here

The library maps Mailchimp's `account_name` to `owner_name` and gates it as contact detail. This report treats `account_name` as the account's name — a company in most cases — and keeps sending it.

So with the switch off, the same value is withheld by one lane and sent by the other. I left it alone because I think the library's mapping is the thing that looks wrong rather than this one, and changing what the report sends about the account is a bigger call than making the switch true. Worth settling before both are in front of merchants.

Also merged `develop-2.4` in to pick the switch up. Still no force-push — `7ace3042` remains an ancestor.

agustinbv:
## Re-reviewed at `51946c3c`

**Your correction on the mechanism is right, and mine was wrong.** I read `cleanAndWarmDefaultScopeData()` and concluded nothing refreshed the in-process array — I never followed into `readData()`, which opens with a full reassignment:

```php
private function readData(): array
{
    $this->data = $this->reader->read();
```

And the `$loadAction` in that path is `function () { return false; }`, so the collector always runs and the refresh always happens on the cache-enabled path. "The ack queue never actually empties" was too strong: it holds only where the config cache is disabled, not on a default install. Your measured table is better evidence than either of our code readings, and the `$cleaner(); return;` branch is exactly the gate.

The fix landing anyway is the right call — the correct behaviour shouldn't depend on how an install has its cache configured.

**Verified on this revision:**

- `.gitattributes` restored byte-for-byte; diff against `acdfbc89` is back to 12 files, 2230 insertions, 0 deletions.
- `handle($storeId, $token, array $pingData, array $stillPending = [])` — the read-back is gone.
- Your reflog diagnosis is better than my "rebase artifact" guess. Reset-onto-base-and-recommit also explains the `PixelStateWriterTest` failures, which I'd assumed were genuinely pre-existing.

I have **not** re-run the suite on this revision, so the 41/64 and 116/213 figures are yours, not independently confirmed by me. Flagging that rather than implying otherwise.

**On 5, 6 and 7** — agreed on all three, including that `Retry-After` is better handled by fixing the claim than by adding machinery that runs hourly regardless. Happy to see 5 split out; the unbounded growth of `mailchimp_errors` is the part I'd weight most, since the query plan only matters once the table is large.

Nothing further blocking from me.

agustinbv:
## On the contact opt-out (`d1186082`)

The gating itself looks right — `accountBlock()` is shared by both the register and ping paths, so one check covers both, and `(bool)'0'` correctly reads a stored "No".

One problem with the claim around it. The new constant's docblock says:

> Shared with the API library, which reads the same path for the same purpose: one switch in the admin has to govern both, or it tells the merchant something that is not true.

**No released version of the library reads that path.** `composer.json` still requires `ebizmarts/mailchimp-lib: ">=3.0.44"`, unchanged in this PR, and:

- `CONTACT_CONFIG = 'mailchimp/telemetry/share_contact'` exists in `src/Mailchimp/Telemetry.php` only on `origin/develop` — `git branch -r --contains` returns that branch and nothing else.
- Tags `3.0.44`, `3.0.45` and `3.0.46` all ship `Telemetry.php`, and none of them contain the constant. `3.0.46` is the newest tag.

Since a Composer install resolves `>=3.0.44` to a tag, every install today gets a library that ignores the switch. So the admin field currently governs the beacon only, while the library's own telemetry keeps sending contact details — which is the exact outcome the docblock says must not happen.

The admin comment makes a stronger promise than the code can keep right now:

> Set this to No to leave the account owner's name and email address out of those reports; everything else keeps working.

Two ways out, depending on sequencing:

1. Cut a `mailchimp-lib` release from `develop` containing `CONTACT_CONFIG`, then raise the constraint to `>=` that version. The promise then holds on any install that satisfies composer.
2. If the lib release isn't imminent, soften the docblock and the admin comment to say the switch governs the extension's diagnostics, and tighten it when the lib ships.

Worth deciding before this merges, since the field's wording is merchant-facing and a privacy switch that only half applies is worse than a narrower one that is accurate.

Separately — `account_name` still travels unconditionally. Usually the company, but Mailchimp accounts created by an individual often carry a person's name there, while `username` is gated. Probably fine, but it's the one remaining field where the boundary between "identity" and "contact details" is fuzzy.

---

Verified at `d1186082` against fresh clones of both repos. I have not run the suite on this revision, and I have not checked whether anything outside `Telemetry.php` in the library reads the path.

agustinbv:
## Re-verification at `d1186082`

Went back over the earlier findings against this revision, independently rather than taking the summary on trust. All three are fixed, and — more to the point — the tests genuinely guard them.

### The fixes

- **Ack queue.** `rememberForAck()` no longer calls `getConfigValue` at all; it takes `array $stillPending = []` and merges from that. The read-back is gone from the class, so the cache-disabled path can't reintroduce it. Passing `[]` unconditionally at the call site is right in both branches, since `handle()` is only reached after a successful ping — the queue is either already empty or was just cleared.
- **Empty subject.** `deliver()` returns `false` on an empty title and the loop `continue`s before the ack.
- **Cron slot.** `$this->slotClaimed` is set on both the already-exists and just-wrote paths, so one claim per process.

### The tests hold the fixes in place

Passing tests prove nothing on their own, so I reverted each model file individually against the current suite:

**`Model/Edge/NotificationDelivery.php` → `981c0304`** — 5 failures, all on the guards:

```
testAcknowledgedUidsAreNotResurrectedByANewDelivery      (BeaconTest)
testWhatIsStillPendingIsNeverReadBackFromConfig
testDeliveredUidsAppendToWhatIsStillPending
testAnEmptySubjectIsNeitherDeliveredNorConfirmed
testAnEmptySubjectDoesNotStopTheRestOfTheBatch
```

**`Model/Edge/Beacon.php` → `981c0304`** — 2 failures:

```
testTheCronSlotIsClaimedOncePerProcessAcrossStoreViews
testDecliningTheContactSwitchLeavesTheOwnerOutOfTheReport
```

`testAcknowledgedUidsAreNotResurrectedByANewDelivery` is the one that matters most — it drives `Beacon::ping()` through a real `NotificationDelivery`, which is exactly the seam nothing covered before. And `testDeliveredUidsAppendToWhatIsStillPending`, which previously asserted the buggy outcome, now fails against the old code. That's the right shape.

### Suite

- Edge suite: **44 tests, 77 assertions, 0 failures**
- Full module unit suite: **119 tests, 226 assertions, 0 failures**

Above your 41/64 and 116/213, which tracks — the opt-out commit landed after those numbers. PHP 8.4.21, PHPUnit 10.5.63. The four `PixelStateWriterTest` failures from the original description are gone, which confirms your read that they were base-branch staleness rather than anything real.

Diff against `acdfbc89` is 14 files, 2399 insertions, **0 deletions**.

### Still open

Nothing new from me. The three you deferred stand as agreed — `getLastError`'s sort, the unused `getRetryAfter()`, the global inbox — plus the `mailchimp-lib` constraint from my previous comment, which is the only one I'd want resolved before merge, since it's the one with merchant-facing wording attached to it.

gonzaloebiz:
## On the library constraint

Your facts are right, and the situation is cleaner than either of us put it. The conclusion drawn from them is the part I'd push back on.

**Confirmed, and stronger than stated.** It is not that the tags ship `Telemetry.php` without `CONTACT_CONFIG` — **no tag ships `Telemetry.php` at all.** Checked the last twelve:

```
3.0.46: no   3.0.45: no   3.0.44: no   3.0.43: no
3.0.42: no   3.0.41: no   3.0.40: no   3.0.39: no
3.0.38: no   3.0.37: no   3.0.36: no   3.0.35: no
```

The file exists only on `develop` and the feature branches, exactly as your `git branch -r --contains` showed for the constant.

**Which is why the outcome you describe cannot happen.** You wrote:

> the library's own telemetry keeps sending contact details — which is the exact outcome the docblock says must not happen

On a released install there is no library telemetry to keep sending anything. And it can never diverge, because the constant entered in **the same commit that created the file** (`b31c5a28` on the library's `develop`):

```
218:+    const CONTACT_CONFIG = 'mailchimp/telemetry/share_contact';
621:+            $value = $this->_helper->getConfigValue(self::CONTACT_CONFIG);
```

The reader and the sender shipped together. There is no version that reports while ignoring the switch, and there cannot be one. So there is no privacy gap today, and none opens later.

What you did correctly identify is an **accuracy problem in my wording**: the docblock asserted in the present tense that the library reads the path, when no installed library does. Corrected in `eda2e9c4` — it now says the library reads it from the release that first gives it anything to report, and that until then this extension's reports are the only thing the switch has to govern. True in both states, so it does not need revisiting when the library ships.

**Sequencing, to be explicit about it:** the library's reporting is not available until its next release, which will be **3.0.47**, and the release of this extension that carries this branch will require that version rather than `>=3.0.44`. Raising the constraint before the tag exists would just leave Composer unable to resolve, so it follows the tag rather than preceding it.

**If you want to review the reporting half now**, point the library at its `develop` branch — that is where `Telemetry.php` lives until 3.0.47 is cut:

```
composer require ebizmarts/mailchimp-lib:dev-develop --prefer-source
```

That is the only way to exercise it today, and worth knowing before concluding the switch does nothing.

I have left the admin field's wording alone. It came from the merged opt-out PR and says the switch keeps the owner's name and address out of "those reports" — which is accurate for whatever set of reports exists on a given install, and will stay accurate as the library joins.

On `account_name` — agreed it is the fuzzy one, and agreed it is probably fine. I raised the same thing from the other side: the library maps it to `owner_name` and gates it, so with the switch off the same value is withheld by one lane and sent by the other. I still think the library's mapping is the thing to settle rather than this, but it should be settled.

Thanks for re-running the reverts independently rather than taking my numbers — that is the part that actually establishes the tests hold.

gonzaloebiz:
## Point 7 — the 429 claim, corrected rather than implemented (`e958a779`)

You were right that it oversold. `Retry-After` is read, logged and carried on the response, and then nothing consults it — `getRetryAfter()` still has no callers and the rate-limited branch just stops. Calling that "backing off" describes scheduling the code does not do.

Taking the option we agreed on: fix the claim, don't add machinery for something that runs hourly anyway.

- The class docblock no longer says the service "asked us to back off". A 429 now reads as the service refusing for now, and the docblock states outright that 429 and the failure case lead to the same action — stop and leave everything alone — and are separate only because they mean different things in the log.
- `isRateLimited()` describes a refusal rather than a back-off.
- `getRetryAfter()` says explicitly that it is **recorded and logged, not obeyed**, that the next attempt is the next hourly tick whatever the header said, and — your second half, which I had not addressed — that a value asking for longer than an hour is therefore ignored. That limitation is now written down instead of implied.

I kept the value on the response rather than deleting the accessor. The log already has it, so deleting would lose nothing today, but the response is where it belongs if honouring it ever becomes worth doing, and the docblock no longer lets it read as though it already is.

The PR description carried the same claim; corrected there too.

**Point 4 is still open and I owe you an answer on it.** I said your reasoning came out stronger and then did not say what we would do, which left it hanging. Looking at it properly next rather than closing it with a one-liner.

119 tests, 226 assertions, green. Current diff against `7710b72f` is 12 files, 2405 insertions, 0 deletions.

gonzaloebiz:
## Point 4 — measured, and fixed (`b80e2c65`)

I owed you an answer here rather than just agreeing it was worse than you said. Measured on a 2.4.8 install before deciding.

### What the 401 path actually costs

First, a correction to the shape. You described it as "a clear-flush plus a re-register-flush per store view per hour". The two flushes are real but they are **not hourly and not in the same tick**: the 401 clears the token in tick N, and the re-registration happens in tick N+1, an hour later. Then it returns to steady state, which writes nothing. A failing re-registration costs nothing at all — `register()` returns null without writing. So an expiry costs two flushes per view in total, spread over two ticks, not a recurring hourly drip.

That makes it smaller than "per hour" implies. What makes it bigger than either of us said is how it scales.

### It is quadratic

`prepareDataForCache()` serialises **and encrypts each scope separately**, so its cost is linear in the number of store views — and the job does one write per view. Measured, holding the rest of the config constant and varying only the store-scope count:

| store views | `prepareDataForCache` | expiry tick (N writes) |
|---|---|---|
| 1 | 2.7 ms | ~0 |
| 10 | 7.0 ms | 0.07 s |
| 25 | 17.3 ms | 0.43 s |
| 50 | 32.1 ms | 1.61 s |
| 100 | 64.0 ms | 6.40 s |

**That table understates it**, and I want to be explicit rather than quote the flattering number: it measures one component. A full `saveConfigValue` end to end measured **28.8 ms** at one store view, against 2.7 ms for `prepareDataForCache` — so roughly 26 ms of other work per write (`readData()` against the database, the cache save), which also grows with the scope count. At fifty views the real expiry tick is well above 1.6 s, under a lock.

For scale: this install's full config is 957 KB serialised, of which the `default` scope is 106 KB.

### The fix

The pattern already existed in the helper and was just not reachable from outside it — `saveMCMinSyncing()` writes through the private atomic upsert and flushes once. The beacon now does the same: writes go through a non-flushing path, and the cache is refreshed **once at the end of the run**, only if anything was written. An expiry wave costs one flush regardless of N. Steady state still writes nothing and flushes nothing.

Deferring is safe here specifically because of the earlier fixes — nothing in the run reads back what it just wrote:

- the token is read before any write for that view;
- pending acks are read before they are cleared;
- the cron slot is settled by the process flag rather than by re-reading.

Those were the three candidates and none depends on it. Had you not found (1) and (3), this change would have been unsafe.

`Helper\Data` gains two public methods, `saveConfigValueWithoutCacheFlush()` and `flushConfigCache()`. That is new API surface on a public extension, which I would rather flag than slip past — the alternative was leaving the beacon to flush per write.

Two tests: five store views whose tokens have all expired produce five writes and **exactly one flush**, and a run that writes nothing does not flush at all. Both fail against the code before this change.

**121 tests, 228 assertions, green.**

One thing I deliberately left alone: `NotificationDelivery` still flushes on its own write. It writes only when a notification actually reached the inbox, which is rare, and it is not part of the expiry wave. Worth revisiting only if notification volume ever makes it matter.

gonzaloebiz:
## `account_name` — you were right, it is gated now (`76f12793`)

I had kept it out of the gate on the reasoning that it names the *account* rather than the owner, and the admin field talks about the owner's name and address. Your counter is the one that decides it: accounts opened by an individual commonly carry that person's name there, while a business account carries a company. A field that may or may not name someone has to be treated as though it does, once the merchant has asked us not to send it.

It also removed a contradiction I had left in place and flagged rather than fixed. The API library reads the same switch and withholds this same value, mapped to `owner_name`. Leaving it ungated here meant one control in the admin produced opposite results depending on which lane was reporting — which is exactly the failure the shared config path exists to prevent. Withholding it in both is the only reading where the switch means one thing.

`account_id`, the plan and the subscriber count still travel, so an opted-out account stays countable and joinable on the receiving side. What is lost is a human-readable name on a dashboard, which is what was declined.

The existing opt-out test now covers `account_name` and fails against the previous code (`account_name was sent after the merchant opted out`). **121 tests, 229 assertions, green.**

## Review point 6 is done too, in its own PR

#2345, stacked on this branch because `NotificationDelivery` does not exist on `develop-2.4` yet. It can be retargeted once this merges.

Worth surfacing one finding from it here, since it bears on the "degrades to a no-op" claim you verified in this PR: **`NotifierInterface` can never deduplicate.** `Inbox::add()` always sets an `internal` key, and `Inbox::parse()` skips its duplicate lookup whenever that key is set — five identical `addNotice()` calls produce five rows. Passing `null` to reach the other branch fails with `Unknown column 'internal'`, because `parse()` only unsets the key on the branch it did not take. So the inbox-side check that would have been the stronger fix is not reachable without giving up the graceful degradation, and the dedupe is per-process instead. That limit is documented rather than papered over.

## Where this leaves your review

Of the seven numbered observations plus the blocker, all are closed except the library constraint, which follows the 3.0.47 tag rather than preceding it — raising it now would only leave Composer unable to resolve.

Two of the follow-ups you agreed to split out are filed with what we found while writing them up. On point 5 in particular, one of your premises needs correcting: **the pruning does exist** — `Cron/ErrorsClean.php` runs hourly as `ebizmarts_clean_errors` and calls `deleteByStorePeriod()`. Your conclusion was right anyway, for a different reason: it is gated on `clean_errors_months > 0`, that path has no default in `config.xml`, and a stored `0` turns out to be written explicitly by saving the config section rather than being absent — so a default would not reach the installs that need it. Worth knowing before anyone attempts the obvious fix.

agustinbv:
## Re-check at `eda9f11a`

Went through all outstanding items. Most are resolved. Two things to raise: **the suite does not pass**, and one of my earlier findings was wrong and I'm withdrawing it.

### Resolved

`.gitattributes` (0 deletions again), the ack queue (`getConfigValue` no longer appears anywhere in `NotificationDelivery.php`), the empty subject, the cron slot, the flush storm, the inbox dedupe, `Retry-After`, and `account_name`.

I checked the `Inbox::parse` reasoning behind the per-process dedupe rather than taking it on trust, and it holds exactly as described: `add()` always sets `internal`, and `parse()` only `unset`s it on the branch that *skips* the duplicate lookup — so `NotifierInterface` genuinely cannot deduplicate, and the `Unknown column 'internal'` on the other branch is real. Per-process was the right call.

`getLastError`'s sort and the unpruned `mailchimp_errors` are untouched, as agreed.

### Withdrawing my `mailchimp-lib` finding — it was wrong

I claimed an install on an older lib would get a switch that governs this extension while the library kept sending contact details. That is not true, and your revised docblock is the accurate description.

`src/Mailchimp/Telemetry.php` **does not exist at `3.0.46`** — `git ls-tree -r --name-only 3.0.46 | grep -i telemetry` returns nothing, and the same holds for every tag. The file, `CONTACT_CONFIG` and the reporting all arrived together in `b31c5a2`, which is unreleased. So no released library can send contact details at all, and there was never a window where the switch described only half of what leaves the server.

My "Telemetry.php exists at 3.0.44/45/46" came from a broken check on my side — the file existed in my working tree and my shell test picked that up instead of the tag. Apologies for the noise; the conclusion you reached was better than mine.

### The suite is red

```
Edge:        49 tests, 82 assertions, 1 error
Full module: 124 tests, 231 assertions, 1 error
```

PHP 8.4.21, PHPUnit 10.5.63, against a 2.4.8 vendor tree.

```
BeaconTest::testTheConfigCacheIsFlushedOncePerRunNotOncePerWrite
Error: Call to a member function info() on null
  Model/Edge/Beacon.php:340   (accountBlock)
  Model/Edge/Beacon.php:241   (ping)
  Model/Edge/Beacon.php:162   (processStore)
  Model/Edge/Beacon.php:124   (execute)
```

The fixture builds store views 1–5 with `https://view{N}.example.com/` and stubs `getGmtDate` to `'0'`. Slot per view:

```
store 1 -> 7      store 2 -> 7      store 3 -> 0   <—
store 4 -> 6      store 5 -> 5
```

Store 3 lands on slot 0, so `shouldSendAccountBlock()` is true at hour 0 and `ping()` calls `accountBlock()`. The helper mock never stubs `getApi()`, so it returns null. `crc32` is deterministic, so this reproduces every run — I ran it repeatedly.

I bisected it: present at `b80e2c65`, `a3c0ece1`, `76f12793` and `eda9f11a`. It has never passed since the test was added, which doesn't match the 121/228 and 124/231 green results in the commit messages. My test *counts* match yours exactly (124/231) — only the outcome differs, so I suspect something environmental on one side rather than a disagreement about the code. Worth reconciling before merge.

One caveat on my own method: my first bisect gave identical counts at every commit, because `git archive` honours the restored `.gitattributes` and strips `Test/`. I redid it with worktrees and the counts then moved as expected (46/49/46/49). The second run is the one I trust.

Fix is presumably a `getApi()` stub in that test, or fixture URLs chosen off slot 0 — though pinning the fixture to avoid the branch would leave the account path untested there, so the stub seems better.

### New: `\Error` escapes every catch, and now skips the flush

The crash above points at something real in production code, not just the test.

All three catches in `Beacon.php` — lines 125, 341 and 469 — are `catch (\Exception)`. What is thrown here is an `\Error`, which none of them catch. It escapes `accountBlock()`, escapes `processStore()`, escapes the per-store handler in `execute()`, and aborts the run for every remaining store view.

That matters more since the flush moved to after the loop. An `\Error` from one store view now skips `flushConfigCache()` entirely, so token clears sit written in `core_config_data` while the config cache keeps serving the old values until something else flushes — the deferred-write design assumes the run reaches its end.

In production `getApi()` returns an injected instance, so a null there is unlikely; the general case is any `TypeError` or similar out of the API path. Catching `\Throwable` in `execute()`, or putting the flush in a `finally`, would close it.

gonzaloebiz:
## You are right, and my green was the false one (`1ad30ea2`)

I reproduced the difference and found the mechanism. This is a real defect in production code, and my passing runs were an artefact.

### Why it passed here and failed there

`getApi()` has no return type, so the helper double hands back `null`, and `accountBlock()` evaluates `$api->root->info()`. What happens next depends on which error handler is installed:

```
--- without Magento's unit error handler ---
Warning: Attempt to read property "root" on null
escaped catch(Exception) -> Error: Call to a member function info() on null

--- with it ---
caught as Exception: PHPUnit\Framework\Exception -> Warning: Attempt to read property "root" on null
```

`dev/tests/unit/framework/bootstrap.php` installs a handler that converts warnings into `PHPUnit\Framework\Exception`. That is an `\Exception`, so the *warning* on `$api->root` is thrown and swallowed by `accountBlock()`'s catch **before** the `\Error` on `->info()` can happen. The test then passes.

So the green result depended on which error handler happened to be installed, not on the code being correct. My 121/228 and 124/231 figures were real runs, and they were worth exactly nothing on this point. Your run is the one that describes the code.

### The production defect is the important half

Fixed at all three sites plus the flush:

- **`\Throwable` alongside `\Exception`** at lines 125, 341 and 469. A `TypeError` out of the API path escaped the account read, escaped `processStore()`, escaped the per-store handler and ended the run for every store view after it.
- **The flush is now in a `finally`.** You identified the sharper consequence: the deferred writes are already in the database, so a run that ends early leaves a cleared token stored while the config cache keeps serving the old one, until something unrelated flushes. The deferred-write design assumed the run reaches its end and nothing guaranteed it. `finally` guarantees it.

### The test

Stubbed `getApi()`, as you suggested — pinning the fixture off slot 0 would have hidden the account path rather than tested it. Two regression tests added: an `Error` in one store view leaves the others still pinged, and a run cut short still flushes what it wrote. Both fail against the previous code (`the config cache was left stale after an aborted run`).

**126 tests, 235 assertions, green here** — and this time that claim rests on the account path being stubbed rather than on an error handler.

### On the library finding

No apology needed, and the withdrawal is worth more than the finding was. Your broken check is the same shape as the one that produced my `git show <tag>:file` result earlier in this work — a pipeline exit code masking a `fatal:`, so the check reported success while reading the working tree. Easy to do and hard to see.

### On your bisect caveat

Worth flagging that you caught it: `git archive` honouring the restored `.gitattributes` and stripping `Test/` would have given identical counts at every commit and a confident wrong answer. That is a nastier trap than the bug it was hunting, because the output looks like evidence.

agustinbv:
## Confirmed at `1ad30ea2`

Reproduced your error-handler diagnosis rather than taking it on trust, and it is exactly right. Built a bootstrap that mimics `dev/tests/unit/framework/bootstrap.php` — warnings converted to `PHPUnit\Framework\Exception` — and ran the **pre-fix** code under both:

```
strict harness (no handler):        Error: Call to a member function info() on null   -> errors
with warning->exception handler:    OK (1 test, 1 assertion)                          -> passes
```

Same code, opposite results. The warning on `$api->root` becomes an `\Exception` and `accountBlock()`'s `catch (\Exception)` swallows it before the `\Error` on `->info()` is ever reached. Your runs were real; they just could not see it.

That generalises past this PR, and is the part I would carry forward: running these tests through Magento's bootstrap means any `catch (\Exception)` in production code can absorb a PHP warning during tests that would be an `\Error` in production. Anything of this shape elsewhere in the module is invisible under the standard invocation.

### The fix verifies

- `catch (\Throwable)` alongside `\Exception` at all three sites (126, 358, 490).
- Flush in `finally`.
- Still additive: 12 files, +2858, −0.

**126 tests, 235 assertions, green in the strict harness** — the environment that caught the original error, so this result is not resting on the handler. Matches your numbers.

Both new tests hold the fix in place. Reverting only `Model/Edge/Beacon.php` to `eda9f11a` and keeping the tests:

```
testAnErrorInOneStoreViewDoesNotAbortTheRest    -> error
testTheFlushStillHappensWhenTheRunIsCutShort    -> failure
```

### One result I discarded

Running the fixed code under my handler gave 16 errors, which looked alarming for a moment. All 16 are one unrelated `Magento\Framework\Registry::registry` deprecation in pre-existing tests, thrown because my handler was blunter than Magento's and did not filter `E_DEPRECATED`. My artefact, not a finding — noting it so the number does not resurface as evidence of something.

### Status

Everything raised in this review is resolved except `getLastError`'s unsupported `ORDER BY id DESC` against an unpruned `mailchimp_errors`, which you split out deliberately. My `mailchimp-lib` finding was withdrawn as incorrect.

Nothing blocking from me.

agustinbv (approved):

PR #2346, opened 2026-08-28
Ship a phpunit configuration for this extension's unit suite

gonzaloebiz:
Two files, 28 lines: a `phpunit.xml.dist` at the extension root and its `export-ignore` entry. Invoke from the Magento root:

```
vendor/bin/phpunit -c app/code/Ebizmarts/MailChimp/phpunit.xml.dist
```

The suite could only be run by knowing how to point Magento's own unit configuration at this directory. Now the extension carries that knowledge, so a machine can run it without being told.

**This is a rewrite of the earlier version, which review found to be wrong in three ways. Recording them, because two of them were arguments this PR was asking to be accepted on.**

## What was wrong

**The bare-clone workflow was promised and never run.** `composer install && vendor/bin/phpunit works in a bare clone` was false: a bare clone makes this extension the root package, and Composer only reads `repositories` from the root, so it needs a `repo.magento.com` entry, an `allow-plugins` entry, and the Magento modules the tests reach — `magento/framework` alone provides none of them.

None of that is worth adding, because the workflow is not real. **This extension only runs inside a Magento project, and a project brings every core module with it.** The right response was to drop the promise, not to satisfy it.

**"Nothing here stubs them" was wrong in the other direction.** `FactoryGenerator` builds `create()` with `'body' => ''`, so it emits:

```php
class CurlFactory { public function create(array $data = []) {} }
```

A stub returning null — not the DI-compiled factory, which takes an `ObjectManagerInterface` and delegates. The reason given for preferring it over a hand-written stub applied to it as written.

**And the verification proved nothing.** `isFactory()` matches on the name ending in `Factory` and never checks that the source class exists, so any name resolves — `Totally\InventedFactory` resolves. Meanwhile no test on this branch reaches a generated class at all: delete the temp directory, run the suite, and it is not recreated. The check could not have distinguished a working generator from an unused one.

## What that leaves

The custom bootstrap, the phpunit dev dependency and the `.gitignore` are gone, and both remaining problems went with them:

- there is no bootstrap of ours choosing between autoloaders, which is what made a local `vendor/` inside `app/code` fatal on duplicate component registration;
- there is no generator of ours to justify. The configuration points at `dev/tests/unit/framework/bootstrap.php`, so the suite sees exactly what the platform's unit tests see — autoloader, generated-class autoloader and error handler. Whether Magento's generator emits a stub is then Magento's decision, not a claim of ours.

## Verification, this time where it matters

The factory question is real, just not on this branch. `Ebizmarts\MailChimp\Model\HTTP\Client\CurlFactory` is referenced nowhere on `develop-2.4`; on the beacon branch, `Model/Edge/Client.php` type-hints it and `ClientTest` mocks it.

So, on **that** branch, with the project's `generated/code/Ebizmarts` moved aside:

- **126 tests, 235 assertions**, green;
- `CurlFactory.php` is generated into `dev/tests/unit/tmp/generated/code/...` — the generator genuinely doing work, on a test that genuinely needs it.

On this branch: **75 tests, 149 assertions**, and nothing needs a factory — which is precisely why the earlier verification could not tell the difference.

## Unchanged

`git archive` still excludes `Test/`, `.gitattributes` and now `phpunit.xml.dist`, so release archives and Composer dist installs are untouched. The existing path — `vendor/bin/phpunit -c dev/tests/unit/phpunit.xml.dist app/code/Ebizmarts/MailChimp/Test/Unit` — is unaffected.

## What this does not do

CI still needs a Magento project, with Marketplace credentials. That was always true and this does not change it. What it changes is that once the project exists, running this extension's suite is one line rather than tribal knowledge.


agustinbv:
Reviewed at `9c867fd2`, in a Magento 2.4.8 / PHP 8.4 environment. Every claim below was run, not read.

**Nothing here breaks the shipped extension.** `git archive` confirms `Test/`, `phpunit.xml.dist`, `.gitignore` and `.gitattributes` are all excluded (400 files, none of the new ones), so release tarballs and Composer dist installs are untouched, and `require-dev` is inert for consumers. The pre-existing path is intact too — `vendor/bin/phpunit -c dev/tests/unit/phpunit.xml.dist app/code/Ebizmarts/MailChimp/Test/Unit` gives **75 tests, 149 assertions**.

What breaks is the thing the PR exists to deliver.

## 1. `composer install && vendor/bin/phpunit` does not work in a bare clone

Three separate blockers, each reproduced from a clean clone:

1. **`magento/framework` is not on Packagist**, and `composer.json` adds no `repositories` entry. Install stops at the first step:
   ```
   Root composer.json requires magento/framework, it could not be found in any version,
   there may be a typo in the package name.
   ```
   Getting past this needs a `repo.magento.com` entry *and* auth keys. Neither is added, and neither is mentioned.

2. After adding that repository, install still aborts. `magento/composer-dependency-version-audit-plugin` arrives transitively via `magento/framework` and is blocked because `composer.json` declares no `config.allow-plugins`:
   ```
   magento/composer-dependency-version-audit-plugin contains a Composer plugin which is
   blocked by your allow-plugins config.
   ```
   This is a hard failure under `--no-interaction`, so it is a hard failure in CI — which is the case the PR is written for.

3. With both fixed, the suite runs and gives **74 errors out of 75 tests**:
   ```
   Tests: 75, Assertions: 1, Errors: 74.
   ```
   `magento/framework` alone does not provide the classes these tests mock. The distinct misses are `Magento\Store\Model\Store`, `Store\Model\ScopeInterface`, `Quote\Model\Quote`, `Quote\Model\Quote\Item`, `Quote\Model\Quote\Item\Option`, `Quote\Model\Quote\Address`, `Catalog\Model\Product`, `Customer\Model\Session`, `Sales\Model\Order`, `Sales\Model\Order\Item`, and `Search\Model\Query`.

The 75 tests / 149 assertions in the description are real, but they come from the app/code-inside-a-Magento-project path — which I confirmed works, `vendor/bin/phpunit -c app/code/Ebizmarts/MailChimp/phpunit.xml.dist` from the Magento root gives `OK (75 tests, 149 assertions)`. The bare-clone claim looks untested. Worth noting the verification section itself describes moving *the project's* `generated/code/Ebizmarts` aside, which only makes sense inside a project.

**A fix does work.** Adding these to `require-dev` gets a bare clone to `OK (75 tests, 149 assertions)`:

```json
"magento/module-store": "*",
"magento/module-quote": "*",
"magento/module-catalog": "*",
"magento/module-customer": "*",
"magento/module-sales": "*",
"magento/module-search": "*"
```

Resolved as 101.1.9 / 101.2.9 / 104.0.9 / 103.0.9 / 103.0.9 / 101.1.9, for 187 packages installed. The `repositories` entry and `config.allow-plugins` are still needed on top of that.

## 2. A new fatal, hit by following the PR's own instructions

Clone into `app/code`, run `composer install` there to get PHPUnit — exactly what the PR says to do — then run the suite from the Magento root, and bootstrap dies before a single test:

```
Error in bootstrap script: LogicException:
Module 'Ebizmarts_MailChimp' from '/var/www/html/app/code/Ebizmarts/MailChimp'
has been already defined in '/var/www/html/app/code/Ebizmarts/MailChimp'.
#0 .../MailChimp/registration.php(12): Magento\Framework\Component\ComponentRegistrar::register(...)
#1 .../MailChimp/vendor/composer/autoload_real.php(41)
#4 .../MailChimp/Test/bootstrap.php(37): require_once('/var/www/html/a...')
```

`Test/bootstrap.php` tries the module's own `vendor/autoload.php` first. By then PHPUnit's launcher has already loaded the project's autoloader, which runs `app/etc/NonComposerComponentRegistration.php` and globs `app/code/*/*/registration.php`. The module's own `"files": ["registration.php"]` then registers a second time, and `ComponentRegistrar::register()` throws on a duplicate component name **even when the path is identical** — it only checks `isset($paths[$type][$name])`.

So the two "ways in" the bootstrap comment describes are not independent: having a local `vendor/` while sitting in `app/code` is fatal, not merely redundant. Preferring the project autoloader over the local one when both are present would avoid it.

## 3. Two claims in the description do not hold

- `Ebizmarts\MailChimp\Model\HTTP\Client\CurlFactory` **does not exist anywhere in this repo**. The only `CurlFactory` reference is `\Magento\Framework\HTTP\Client\CurlFactory`, in `Model/Api/Result.php` lines 47, 58 and 67 — a Magento class, not an Ebizmarts one.

- > **Nothing here stubs them.** […] It produces the *real* generated class on demand […] A hand-written stub would drift from what the compiler emits

  `Magento\Framework\TestFramework\Unit\Autoloader\FactoryGenerator` is documented in its own docblock as "Generates a simple factory class with create() method", and builds that method with `'body' => ''`. What it actually wrote to disk when I loaded `CurlFactory` through this bootstrap:

  ```php
  <?php
  namespace Magento\Framework\HTTP\Client;

  class CurlFactory
  {
      public function create(array $data = [])
      {
      }
  }
  ```

  An empty `create()` returning `null`, matched purely by the `<X>Factory` name convention — `isFactory()` never checks that the source class exists. The DI-compiled factory takes an `ObjectManagerInterface` and delegates to `$this->_objectManager->create($this->_instanceName, $data)`. This *is* a stub, just Magento's own rather than a hand-written one. That is a perfectly good reason to use it, but it is the opposite of the stated justification, and the "would drift from what the compiler emits" argument applies to it as written.

Related: **the generator is never exercised by this suite.** Delete `Test/tmp` and run the 75 passing tests — nothing is regenerated, the directory is not even created. No test reaches a generated class; `QueryFactory` in `Test/Unit/Block/Pixel/SearchTest.php` is a real hand-written class in `magento/module-search`. That is fine as insurance for future tests, but it means the mechanism the PR is built around is currently carrying no load, and a regression in it would go unnoticed.

## Summary

The `.gitattributes` work is correct and the app/code path genuinely works. The bare-clone workflow in the title and opening line does not, and needs the three fixes in §1 before the claim is true. §2 is a regression introduced by this PR for anyone who follows its instructions inside a Magento project. §3 is a description accuracy issue, but it is the argument the review is being asked to accept, so it seems worth correcting before merge.

gonzaloebiz:
All three sections are correct. I reproduced each one rather than take them, and every claim held. Rewritten at `bf92016d` — the PR is now two files and 28 lines, and both of the arguments you challenged are gone rather than patched.

## §3 — you are right, and the verification was circular

`FactoryGenerator` builds `create()` with `'body' => ''`, so what it writes is a stub returning null. The DI-compiled factory takes an `ObjectManagerInterface`, holds `_instanceName` and delegates. My "produces the real generated class" and "nothing here stubs them" were both false, and "a hand-written stub would drift from what the compiler emits" describes the thing I was defending.

The verification was worse than wrong, it was incapable of being wrong. `isFactory()` matches on the name and never checks the source class exists — I ran it against `Totally\InventedFactory` and it generates that too. And you are right that nothing exercises it: delete the temp directory, run the suite, it is not recreated.

**One correction in the other direction, which explains where the need actually lives.** `Ebizmarts\MailChimp\Model\HTTP\Client\CurlFactory` is referenced nowhere on `develop-2.4`, as you found — but on the beacon branch `Model/Edge/Client.php` type-hints it and `Test/Unit/Model/Edge/ClientTest.php` mocks it. That is the branch you were reviewing when you first hit this and hand-wrote a stub. The problem you originally reported was real; I then built the fix against a branch where it does not occur, which is how a verification that proved nothing could look like it passed.

## §1 — right, and the conclusion is to drop the goal

A bare clone makes this extension the root package, and Composer reads `repositories` only from the root, so your three blockers are real and a `repositories` entry would be inert in the normal path anyway.

But the goal was mine and it was wrong. This extension only runs inside a Magento project, and a project brings every core module with it — so adding your six packages to `require-dev` would have made the suite pass while writing down that the extension needs six modules to test, when the runtime code touches classes from twenty. Fixing that under-declaration is not worth doing either, for the same reason the platform makes it invisible. The promise is removed instead.

## §2 — gone rather than fixed

The custom bootstrap is deleted, so nothing of ours chooses between autoloaders and the duplicate `ComponentRegistrar::register()` cannot happen. I confirmed the mechanism you described: it throws on `isset($paths[$type][$name])` without comparing paths, and this extension's `composer.json` has `"files": ["registration.php"]`.

## What is left, and how it was checked

A `phpunit.xml.dist` pointing at `dev/tests/unit/framework/bootstrap.php`, so the suite sees what the platform's unit tests see. Whether Magento's generator emits a stub is then Magento's decision rather than a claim of mine.

Checked on the branch where a test actually mocks a generated factory — the beacon branch, with `generated/code/Ebizmarts` moved aside: **126 tests, 235 assertions**, and `CurlFactory.php` written into `dev/tests/unit/tmp/generated/code/...`. On `develop-2.4`: **75 tests, 149 assertions**, nothing generated, which is the result the old verification could not distinguish from success.

Thanks for running it rather than reading it. Three of the five things I had asserted here were the kind that sound right and had never been executed.

agustinbv:
Your correction on `CurlFactory` is right, and it deserves finishing, because the fact that completes it is on `develop-2.4` too.

`Model/HTTP/Client/Curl.php` is in this branch — `class Curl implements \Magento\Framework\HTTP\ClientInterface`, injected directly in `Helper/Http.php`. So `Ebizmarts\MailChimp\Model\HTTP\Client\CurlFactory` was always a legitimate class *here*, not only on the beacon branch. `Code\Generator::generateClass()` matches the `<X>Factory` suffix, strips it to derive the source class, and `Code\Generator\Autoloader::load()` is registered as an spl autoloader, so any reference resolves and materialises into `generated/code/`. I ran Magento's real `ObjectManager\Code\Generator\Factory` against it, and it generated:

```php
public function __construct(\Magento\Framework\ObjectManagerInterface $objectManager, $instanceName = '\\Ebizmarts\\MailChimp\\Model\\HTTP\\Client\\Curl')
{ ... }

public function create(array $data = [])
{
    return $this->_objectManager->create($this->_instanceName, $data);
}
```

So "does not exist anywhere in this repo" was a wrong claim rather than loose phrasing. For a generated factory, having no source file *is* the definition — I grepped for the factory name and treated its absence as disproof, when the check that means anything is whether the base class exists. What is genuinely absent on `develop-2.4` is a **reference**, and that distinction is the one your beacon-branch explanation supplies. The contrast with `FactoryGenerator` survives it intact, and is sharper for having the compiler's output to sit next to.

## `bf92016d`

Read and run. Two files, no bootstrap of ours, no composer.json changes, no promise that needs a dependency set behind it. Removing the goal was the right call — the six `require-dev` packages I proposed would have made the suite pass while encoding a dependency claim that is both wrong and unnecessary inside a project.

What I checked:

- **The relative bootstrap resolves.** Four levels up from `app/code/Ebizmarts/MailChimp` is the Magento root; `realpath` on a real 2.4.8 layout gives `/var/www/html/dev/tests/unit/framework/bootstrap.php`, and the file is there.
- **PHPUnit loads it and discovers the suite** — all 75 tests found and executed. I ran it from an isolated root so I will not quote a green count from that harness, but `dev/tests/unit/phpunit.xml.dist` declares `bootstrap="./framework/bootstrap.php"`, the same file, and that gave `OK (75 tests, 149 assertions)` for this suite in my earlier run. `Test/Unit` is untouched by this branch, so it is the same 75.
- **Generated code goes where the platform puts it.** `TESTS_TEMP_DIR` resolves to `dev/tests/unit/tmp`, inside the project, so nothing lands in the module.
- **`.gitattributes`** keeps it out of `git archive` as before.

### One thing it writes into the clone

Every run leaves a `.phpunit.result.cache` in the module root:

```
$ ls -a app/code/Ebizmarts/MailChimp/
.  ..  .phpunit.result.cache  Test  phpunit.xml.dist
```

With no `cacheDirectory` attribute, `TextUI/Configuration/Merger.php` falls through to:

```php
} elseif ($xmlConfiguration->wasLoadedFromFile()) {
    $testResultCacheFile = dirname(realpath($xmlConfiguration->filename())) . DIRECTORY_SEPARATOR . '.phpunit.result.cache';
}
```

which is the config file's own directory. The previous revision covered this with the `.gitignore` it shipped; that went out with everything else, so an app/code clone now gets an untracked file after every run, sitting where it is easy to commit by accident. `cacheResult="false"` on the root element removes it outright, a `cacheDirectory` under the project's temp area redirects it, or one line of `.gitignore` covers it — whichever fits the "one less thing to keep correct" principle you are applying.

### Minor, and by construction

The four-level path ties this to the app/code layout; from `vendor/mailchimp/mc-magento2` it would climb past the Magento root. That is consistent rather than a flaw — the file is `export-ignore`d so a dist install never has it, and the header says where to invoke it from. Worth knowing before anyone relocates it.

Nothing else. §1 and §2 are dissolved rather than patched, which is the better outcome of the two.

gonzaloebiz:
Fixed in `f1cca55f`, and your correction on `CurlFactory` finishes the picture better than mine did.

## The cache file

Reproduced, and I had already seen it without registering it — it showed up as `?? .phpunit.result.cache` in a `git status` earlier in this work and I read past it. It is written into the extension root because `Merger.php` falls back to the configuration file's own directory when no `cacheDirectory` is set, exactly as you quote.

Turned off with `cacheResult="false"` rather than redirected or ignored. Redirecting and ignoring both work, but each is a second thing to keep correct, and the cache buys nothing on a suite that runs in a tenth of a second. The extension is clean after a run now.

## On `CurlFactory` — your correction is the sharper one

You are right that `Model/HTTP/Client/Curl.php` is on `develop-2.4` and injected in `Helper/Http.php`, so `Ebizmarts\MailChimp\Model\HTTP\Client\CurlFactory` was always a legitimate generatable class here. Verified both.

And the distinction you draw is the one that matters: **for a generated factory, having no source file is the definition, not evidence of absence.** What is missing on `develop-2.4` is a *reference*. My original PR text — "has no source file at all, it exists only under `generated/code/`" — read that fact as though it were a finding, when it is simply what a factory looks like.

Which means the contrast you set up survives and is better with the compiler's real output beside it: the DI-compiled factory takes an object manager and delegates to it; `FactoryGenerator` emits `create()` with an empty body. Both are "generated"; only one does anything. That is the whole reason the argument I built on it was wrong.

Worth noting the shape of both of our mistakes here, since they are the same one from opposite sides. You grepped for the factory name and treated its absence as disproof; I ran a check that resolved the factory and treated its presence as proof. Neither test could have failed in the direction that mattered — the class is generated on demand from the *name*, so absence and presence were both guaranteed regardless of the thing we each wanted to know.

## On the four-level path

Agreed, and worth having it written down as you put it: it ties the file to the `app/code` layout, and from `vendor/mailchimp/mc-magento2` it would climb past the Magento root. It is `export-ignore`d so a dist install never receives it, and the header says where to invoke it from — but anyone relocating this file needs to know that is load-bearing.

agustinbv (approved):

PR #2347, opened 2026-08-28
Bound mailchimp_errors with a row ceiling, without overriding anyone

gonzaloebiz:
`mailchimp_errors` grows without a ceiling on a default install. This bounds it, inside the window the cron group allows.

## Why the cleaner that already exists does not

`Cron/ErrorsClean.php` runs hourly and deletes rows older than `clean_errors_months`, but it is gated on `$period > 0`, and that path has **no default in `config.xml`**. So the job runs every hour and deletes nothing.

A `config.xml` default would not fix it. A stored `0` is **explicit, not absent** — it is written in the same timestamp as fourteen other rows of the same group, because saving the section persists every field in it. And that inverts who a default reaches: for the table to grow, ecommerce has to be active; for that, someone saved the group; saving it wrote the `0`. **The population at risk is by construction the one a default would not help.**

Rewriting those zeroes would reach them, but a stored `0` is genuinely ambiguous between "never thought about it" and "keep my errors", and guessing wrong there is invisible.

## What this does instead

Bounds the table rather than arguing about the preference. They answer different questions and both hold:

- `clean_errors_months` — *how long do I want errors kept?* A preference, and forever is a legitimate answer.
- the ceiling — *whatever was chosen, one store view never holds more than `MAX_ROWS_PER_STORE`.* A safety property.

Nobody is overridden, and nothing has to decide what a `0` meant. It runs **unconditionally**, because a store with the age clean switched off is exactly the store that needs bounding.

## Converging without starving the sync

The ceiling has its own delete limit and repeats within a run, so a table far past the ceiling comes down in a tick or two rather than the six weeks a fixed 1,000-rows-an-hour slice would take. Measured: **1,000,000 rows per store view in ~16 s**, no single statement holding locks longer than **265 ms**.

That created a second problem, and it is why the budget exists. `etc/cron_groups.xml` gives this group a `schedule_lifetime` of two minutes and `use_separate_process`, and `ebizmarts_ecommerce` and `ebizmarts_webhooks` are `*/5` jobs in it. A job that overruns does not delay its siblings — Magento marks them **missed**. So the whole job now stops on a wall-clock budget well inside that window, and the next tick continues.

The budget covers **both halves deliberately**. The age-based delete filters on `date_add(added_at, ...)` — a function on the column, so it cannot seek: on a store whose errors are all recent it walks the partition to delete nothing, measured in review at ~1.55 µs per row scanned. The sync it would starve does not care which half spent the window.

And the **ceiling runs first**: the safety property draws on the budget before the preference does, and every row it removes is a row the non-seeking scan does not have to walk.

## The index

`(store_id, id)`, declared in `db_schema.xml` and registered in the whitelist.

An earlier revision argued it could wait. That argument was measured wrong and is retracted in the thread. What it actually buys, stated narrowly:

| | without | with |
|---|---|---|
| skewed, pre-drain (100k old rows under 400k newer) | 19.3 ms | 0.7 ms |
| this PR's offset query, bounded state, MariaDB 10.6 | 0.77 ms | 0.71 ms |
| `ORDER BY id DESC LIMIT 1`, bounded state, 10.6 | 52.49 ms | 0.04 ms |

So in **this** PR it buys a transient win while a table drains. The permanent win is on the `LIMIT 1` shape, which arrives with the beacon. On MariaDB 11.4 the optimiser does not take the backward walk in the bounded state at all — so this is insurance against plan choice that varies by engine version, at a measured **6 µs** per error insert. Single-store-view installs are unaffected either way: the walk only crosses rows belonging to *other* views.

## What it does not cover

Rows left by a **deleted** store view. `store_id` carries no foreign key and `getStores()` returns neither deleted views nor the admin store, so neither cleaner ever visits them. Already true of the age-based clean; stated here because this is the change that claims the table is bounded.

Bytes are bounded only loosely: two `TEXT` columns at 65,535 each put a row under ~128 KB and one store view under ~640 MB — **per view**, so a fifty-view install has a 250,000-row ceiling. Against ~110 bytes an error body measures in practice. If that constant ever bites, the answer is to cap the stored body, not the row count.

The non-seeking scan itself is untouched. Rewriting the predicate to seek needs an index on `added_at` to pay, which is a second index and a second write per insert — a separate change.

## Merchant-facing

The admin field said only that errors are removed after N months. It now also says the ceiling applies whatever is chosen, and that removing errors removes what the Mailchimp columns on the product, customer and order grids show — which is where a merchant meets them.

## Tests

**Ten** in `ErrorsCleanTest`, and the module suite is **136 tests, 245 assertions** green. Each new one was checked failing against the revision before it: the ceiling applies with the age clean off, both run when it is on, a failing age clean does not skip the ceiling, every view is offered it even when one throws, the ceiling keeps going until a pass deletes nothing, one run is bounded by the pass cap, it stops on the clock, the budget covers the age clean too, an exhausted budget stops the loop rather than visiting every view, and the ceiling runs before the age clean.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,3 +5,4 @@
 3. Commits must be accompanied by meaningful commit messages. (include a **closes** in the commit that close the issue)
 4. PRs that include bug fixing must be accompanied by the creation of an issue describing the bug following this [Issue reporting guidelines](https://github.com/mailchimp/mc-magento2/wiki/Issue-reporting-guidelines).
 5. For large features or changes, please open an issue and discuss first. This may prevent duplicate or unnecessary effort, and it may gain you some additional contributors.
+6. PRs that claim a performance change must show the measurement, and the measurement must be taken after `ANALYZE TABLE`. **A newly created index benchmarks worse than no index until the optimiser's statistics catch up**, so the first result you get is often the opposite of the true one — an index that turned out to be 28x faster first measured 2.5x slower, and that number very nearly went into a review. This is not specific to any one table or index.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,4 +5,3 @@
 3. Commits must be accompanied by meaningful commit messages. (include a **closes** in the commit that close the issue)
 4. PRs that include bug fixing must be accompanied by the creation of an issue describing the bug following this [Issue reporting guidelines](https://github.com/mailchimp/mc-magento2/wiki/Issue-reporting-guidelines).
 5. For large features or changes, please open an issue and discuss first. This may prevent duplicate or unnecessary effort, and it may gain you some additional contributors.
-6. PRs that claim a performance change must show the measurement, and the measurement must be taken after `ANALYZE TABLE`. **A newly created index benchmarks worse than no index until the optimiser's statistics catch up**, so the first result you get is often the opposite of the true one — an index that turned out to be 28x faster first measured 2.5x slower, and that number very nearly went into a review. This is not specific to any one table or index.
--- a/Cron/ErrorsClean.php
+++ b/Cron/ErrorsClean.php
@@ -74,8 +74,9 @@ class ErrorsClean
      * keeps every statement small and still converges within a single run.
      *
      * Larger than the age-based limit because each pass also runs the offset
-     * query that finds the cut-off, and on a table far above the ceiling that
-     * query is the expensive half. Fewer, bigger passes means fewer of them.
+     * query that finds the cut-off. That query is served by the
+     * (store_id, id) index, but it is still a query per pass, so fewer and
+     * bigger passes means fewer of them.
      *
      * Up to a million rows per store view per run — measured at about 16
      * seconds of work for that, with no single statement holding locks longer
--- a/Cron/ErrorsClean.php
+++ b/Cron/ErrorsClean.php
@@ -40,15 +40,9 @@ class ErrorsClean
      * This bounds it regardless, without overriding anyone: a merchant who
      * wants errors kept still gets them kept, up to here.
      *
-     * It bounds the store views the extension can see. `store_id` carries no
-     * foreign key and getStores() returns neither deleted views nor the admin
-     * store, so rows left behind by a removed store view are visited by neither
-     * cleaner. That was already true of the age-based clean; it is written down
-     * here because this is the change that claims the table is bounded, and for
-     * those rows it is not. The number is
-     * generous on purpose — the table exists to diagnose recent failures, and
-     * a store erroring hard enough to reach this has a bigger problem than the
-     * one this prevents.
+     * The number is generous on purpose — the table exists to diagnose recent
+     * failures, and a store erroring hard enough to reach this has a bigger
+     * problem than the one this prevents.
      *
      * It bounds bytes too, but loosely, and the distinction is worth stating
      * rather than glossing. `type` and `errors` are TEXT, whose own ceiling is
@@ -59,6 +53,13 @@ class ErrorsClean
      * the real problem; what it leaves is a constant that may or may not be
      * comfortable. If it ever proves not to be, the answer is to cap the
      * stored body rather than the row count.
+     *
+     * It bounds the store views the extension can see, and only those.
+     * `store_id` carries no foreign key and getStores() returns neither deleted
+     * views nor the admin store, so rows left behind by a removed store view
+     * are visited by neither cleaner. That was already true of the age-based
+     * clean; it is written here because this is the change that claims the
+     * table is bounded, and for those rows it is not.
      */
     const MAX_ROWS_PER_STORE = 5000;
 
@@ -88,6 +89,23 @@ class ErrorsClean
     const MAX_PASSES     = 100;
 
     /**
+     * Seconds the ceiling may spend across every store view in one run.
+     *
+     * A pass cap bounds work, not time, and how long a pass takes depends on
+     * row sizes, replication and purge — the things this cannot know. That
+     * matters because cron_groups.xml gives this group a schedule_lifetime of
+     * two minutes and runs it in a single process: a job scheduled while this
+     * is still going does not wait, it is marked missed. The sync and webhook
+     * jobs are in that group and run every five minutes, so a long drain does
+     * not delay them, it skips them.
+     *
+     * Well inside the two-minute window, so the ceiling can never be the reason
+     * a sync did not run. A table far past the ceiling then comes down over two
+     * or three ticks instead of one, which is the cheaper mistake.
+     */
+    const MAX_RUNTIME_SECONDS = 30;
+
+    /**
      * @param \Ebizmarts\MailChimp\Helper\Data $helper
      * @param \Ebizmarts\MailChimp\Model\MailChimpErrors $chimpErrors
      * @param \Magento\Store\Model\StoreManager $storeManager
@@ -102,8 +120,20 @@ class ErrorsClean
         $this->chimpErrors = $chimpErrors;
         $this->storeManager = $storeManager;
     }
+    /**
+     * Seam so the budget can be exercised without a test that waits for it.
+     *
+     * @return float
+     */
+    protected function now()
+    {
+        return microtime(true);
+    }
+
     public function execute()
     {
+        $deadline = $this->now() + self::MAX_RUNTIME_SECONDS;
+
         foreach ($this->storeManager->getStores() as $storeId => $val)
         {
             $period = $this->helper->getConfigValue(\Ebizmarts\MailChimp\Helper\Data::XML_CLEAN_ERROR_MONTHS, $storeId);
@@ -123,6 +153,11 @@ class ErrorsClean
             try {
                 $removed = 0;
                 for ($pass = 0; $pass < self::MAX_PASSES; $pass++) {
+                    if ($this->now() >= $deadline) {
+                        // Out of time rather than out of work. The next tick
+                        // picks up where this one stopped.
+                        break;
+                    }
                     $deleted = $this->chimpErrors->deleteOverflowByStore(
                         $storeId,
                         self::MAX_ROWS_PER_STORE,
--- a/Cron/ErrorsClean.php
+++ b/Cron/ErrorsClean.php
@@ -43,9 +43,14 @@ class ErrorsClean
      * a store erroring hard enough to reach this has a bigger problem than the
      * one this prevents.
      *
-     * Note it bounds rows, not bytes: `type` and `errors` are TEXT, so a store
-     * whose failures carry large bodies still holds more than a store whose do
-     * not.
+     * It bounds bytes too, but loosely, and the distinction is worth stating
+     * rather than glossing. `type` and `errors` are TEXT, whose own ceiling is
+     * 65,535 bytes, so a store view cannot exceed roughly 128 KB per row —
+     * about 640 MB here, against the ~110 bytes an error body actually
+     * measures in practice. What this removes is the unbounded case, which was
+     * the real problem; what it leaves is a constant that may or may not be
+     * comfortable. If it ever proves not to be, the answer is to cap the
+     * stored body rather than the row count.
      */
     const MAX_ROWS_PER_STORE = 5000;
 
--- a/Cron/ErrorsClean.php
+++ b/Cron/ErrorsClean.php
@@ -32,7 +32,7 @@ class ErrorsClean
     /**
      * Rows kept per store view, whatever the merchant asked for.
      *
-     * The age-based clean below answers "how long do I want errors kept?", and
+     * The age-based clean answers "how long do I want errors kept?", and
      * keeping them forever is a legitimate answer — it is also the answer every
      * install carries by default, since the field has no default and saving the
      * configuration section stores a 0. So on its own it bounds nothing.
@@ -108,7 +108,8 @@ class ErrorsClean
      * cannot seek: on a store whose errors are all recent it walks the whole
      * partition to delete nothing. It has always done that, and it is the more
      * expensive half here. The sync it would starve does not care which half
-     * spent the window.
+     * spent the window, which is why the budget covers both and why the ceiling
+     * draws on it first.
      *
      * Spent in getStores() order, so while a drain is running the earlier views
      * are favoured and a later one can get nothing on a given tick. It still
@@ -194,6 +195,17 @@ class ErrorsClean
                 }
             } catch (\Exception $e) {
                 $this->helper->log($e->getMessage());
+            } catch (\Throwable $t) {
+                // An Error is not an Exception, and one store view's failure
+                // must not end the run for the views after it.
+                $this->helper->log($t->getMessage());
+            }
+
+            if ($this->now() >= $deadline) {
+                // The ceiling used the window. One age-based statement can walk
+                // a whole partition, so starting one here would overshoot by
+                // however long that takes; the next tick runs it instead.
+                continue;
             }
 
             $period = $this->helper->getConfigValue(\Ebizmarts\MailChimp\Helper\Data::XML_CLEAN_ERROR_MONTHS, $storeId);
@@ -203,6 +215,8 @@ class ErrorsClean
                     $this->chimpErrors->deleteByStorePeriod($storeId,$period,self::LIMIT);
                 } catch (\Exception $e) {
                     $this->helper->log($e->getMessage());
+                } catch (\Throwable $t) {
+                    $this->helper->log($t->getMessage());
                 }
             }
         }
--- a/Cron/ErrorsClean.php
+++ b/Cron/ErrorsClean.php
@@ -89,7 +89,7 @@ class ErrorsClean
     const MAX_PASSES     = 100;
 
     /**
-     * Seconds the ceiling may spend across every store view in one run.
+     * Seconds the whole job may spend across every store view in one run.
      *
      * A pass cap bounds work, not time, and how long a pass takes depends on
      * row sizes, replication and purge — the things this cannot know. That
@@ -99,9 +99,21 @@ class ErrorsClean
      * jobs are in that group and run every five minutes, so a long drain does
      * not delay them, it skips them.
      *
-     * Well inside the two-minute window, so the ceiling can never be the reason
-     * a sync did not run. A table far past the ceiling then comes down over two
+     * Well inside the two-minute window, so this job can never be the reason a
+     * sync did not run. A table far past the ceiling then comes down over two
      * or three ticks instead of one, which is the cheaper mistake.
+     *
+     * It covers the age-based delete as well as the ceiling, deliberately. That
+     * one filters on `date_add(added_at, ...)`, a function on the column, so it
+     * cannot seek: on a store whose errors are all recent it walks the whole
+     * partition to delete nothing. It has always done that, and it is the more
+     * expensive half here. The sync it would starve does not care which half
+     * spent the window.
+     *
+     * Spent in getStores() order, so while a drain is running the earlier views
+     * are favoured and a later one can get nothing on a given tick. It still
+     * converges, because the drain is finite and the next tick starts where
+     * this one stopped.
      */
     const MAX_RUNTIME_SECONDS = 30;
 
@@ -136,6 +148,16 @@ class ErrorsClean
 
         foreach ($this->storeManager->getStores() as $storeId => $val)
         {
+            if ($this->now() >= $deadline) {
+                // At the top, so the budget covers the whole job rather than
+                // only the ceiling below it — the age-based delete is the more
+                // expensive half on the tables this exists for, and a job that
+                // overruns is one whose siblings get marked missed, whichever
+                // half spent the window. Breaking rather than continuing also
+                // stops walking the remaining views to do nothing.
+                break;
+            }
+
             $period = $this->helper->getConfigValue(\Ebizmarts\MailChimp\Helper\Data::XML_CLEAN_ERROR_MONTHS, $storeId);
             if ($period > 0) {
                 try {
--- a/Cron/ErrorsClean.php
+++ b/Cron/ErrorsClean.php
@@ -158,20 +158,16 @@ class ErrorsClean
                 break;
             }
 
-            $period = $this->helper->getConfigValue(\Ebizmarts\MailChimp\Helper\Data::XML_CLEAN_ERROR_MONTHS, $storeId);
-            if ($period > 0) {
-                try {
-                    $this->helper->log("Cleaning errors for store [$storeId] older than $period months");
-                    $this->chimpErrors->deleteByStorePeriod($storeId,$period,self::LIMIT);
-                } catch (\Exception $e) {
-                    $this->helper->log($e->getMessage());
-                }
-            }
-
-            // Deliberately outside the check above. The age-based clean is a
-            // preference and can legitimately be switched off; the ceiling is
-            // not, and a store that has switched the preference off is exactly
-            // the store that needs it.
+            // The ceiling goes first. It is the safety property, so it gets the
+            // budget before the preference does — and the age-based delete
+            // below cannot seek, so every row the ceiling removes here is a row
+            // that one does not have to scan. Running it second also means it
+            // is never the half that starves the ceiling of passes.
+            //
+            // It also runs unconditionally. The age-based clean is a preference
+            // and can legitimately be switched off; the ceiling is not, and a
+            // store that has switched the preference off is exactly the store
+            // that needs it.
             try {
                 $removed = 0;
                 for ($pass = 0; $pass < self::MAX_PASSES; $pass++) {
@@ -198,6 +194,16 @@ class ErrorsClean
                 }
             } catch (\Exception $e) {
                 $this->helper->log($e->getMessage());
+            }
+
+            $period = $this->helper->getConfigValue(\Ebizmarts\MailChimp\Helper\Data::XML_CLEAN_ERROR_MONTHS, $storeId);
+            if ($period > 0) {
+                try {
+                    $this->helper->log("Cleaning errors for store [$storeId] older than $period months");
+                    $this->chimpErrors->deleteByStorePeriod($storeId,$period,self::LIMIT);
+                } catch (\Exception $e) {
+                    $this->helper->log($e->getMessage());
+                }
             }
         }
     }
--- a/Cron/ErrorsClean.php
+++ b/Cron/ErrorsClean.php
@@ -30,6 +30,26 @@ class ErrorsClean
     const LIMIT = 1000;
 
     /**
+     * Rows kept per store view, whatever the merchant asked for.
+     *
+     * The age-based clean below answers "how long do I want errors kept?", and
+     * keeping them forever is a legitimate answer — it is also the answer every
+     * install carries by default, since the field has no default and saving the
+     * configuration section stores a 0. So on its own it bounds nothing.
+     *
+     * This bounds it regardless, without overriding anyone: a merchant who
+     * wants errors kept still gets them kept, up to here. The number is
+     * generous on purpose — the table exists to diagnose recent failures, and
+     * a store erroring hard enough to reach this has a bigger problem than the
+     * one this prevents.
+     *
+     * Note it bounds rows, not bytes: `type` and `errors` are TEXT, so a store
+     * whose failures carry large bodies still holds more than a store whose do
+     * not.
+     */
+    const MAX_ROWS_PER_STORE = 5000;
+
+    /**
      * @param \Ebizmarts\MailChimp\Helper\Data $helper
      * @param \Ebizmarts\MailChimp\Model\MailChimpErrors $chimpErrors
      * @param \Magento\Store\Model\StoreManager $storeManager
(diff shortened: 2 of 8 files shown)


agustinbv (commented):
Reviewed at `9648048a` against `develop-2.4`, on a fresh clone plus a throwaway MariaDB 11.4 and a PHP 8.4 / Magento 2.4 container.

**Nothing here breaks.** No crash, no SQL defect, no logic error. Four things below are worth weighing before merge, but none of them is a blocker.

## What I verified

- **The drain is correct.** I rebuilt the table from `etc/db_schema.xml` and ran the two statements exactly as the code issues them. 250 rows, ceiling 100, limit 40 → deleted 40, 40, 40, 30, 0; stopped at exactly 100; kept the newest row; left the other store view untouched; no-op under the ceiling. That reproduces the table in the description.
- **`limit(1, $keep)` → `LIMIT 1 OFFSET keep`** (`Magento\Framework\DB\Select::limit($count, $offset)`), so the off-by-one is right: a store holding exactly `keep` rows is untouched, `keep + 1` deletes one.
- **The raw `DELETE` is safe.** Positional `?` binds pass straight through `Pdo\Mysql::_prepareQuery()` — it only rewrites non-int bind keys — the multi-query guard doesn't trip, and `rowCount()` on the returned statement is affected-rows. The identifier comes from `getTable()` and the limit is cast, as the comment says.
- **`fetchOne()` returning `false`** under the ceiling is handled.
- **Tests.** The four new ones pass; the full module suite is 79 tests / 153 assertions green on PHP 8.4. They do error against the code before the change, so "all failing before this" holds.
- **PHPCS `Magento2`:** 0 errors before and after. +5 warnings, all of kinds already present in these files (missing comment blocks, the raw-SQL warning, argument alignment).
- The missing trailing newline in `Cron/ErrorsClean.php` is pre-existing on `develop-2.4`, not introduced here.

The premise checks out too: `etc/config.xml` carries no `clean_errors_months` default (only `clean_support_period`), and `Model/Config/Source/Months` makes `"0" => "No"` the first option.

## Four things worth raising

### 1. The drain is much slower than "several runs", on exactly the tables this targets

`LIMIT = 1000` per run, hourly, is 24k rows/day per store view. A store sitting on 1M rows takes ~41 days; 5M takes ~208. The description's worked example is five runs, which reads as "this resolves shortly after upgrade" — for the installs that motivated the change it doesn't. The ceiling currently reuses the age-clean's `LIMIT`; giving it its own, larger one would close the gap without touching the age-based path.

### 2. The offset query has no supporting index, and it runs forever — not only while draining

The only index is `(store_id, regtype, original_id)`, which is not ordered by `id` within a store, so `ORDER BY id DESC` can't be served from it. Measured with 100k store-1 rows (older ids) and 400k store-2 rows (newer ids):

| plan | result |
| --- | --- |
| as written | `type: index, key: PRIMARY` — a backward PK walk past all 400k store-2 rows — **45 ms** |
| with `KEY (store_id, id)` | `type: range, Using index` — **0.8 ms** (~55×) |

The `SELECT` runs unconditionally for every store view on every hourly run, including once the table is bounded. The stated reason for skipping the index — "with the table bounded the sort is over a few thousand rows" — is circular: that holds only *after* bounding, which per (1) can take months, and it never holds on a multi-store install where another view's rows sit on top of the primary key. The counter-argument about an extra write per error insert is real but small next to an hourly index walk.

### 3. Orphan rows are still unbounded

`getStores()` returns only live store views and excludes the admin store, and `mailchimp_errors.store_id` has no foreign key — `db_schema.xml` declares the primary key and that one index, nothing else. Delete a store view and its error rows are never visited again by either cleaner. Pre-existing for the age-based clean, but this PR is the one making the unqualified claim that the table no longer grows without bound, so it's worth qualifying.

### 4. "No" in the admin field stops meaning "never delete"

`Model/Config/Source/Months` offers `No / 1 / 2 / 3 / 4`, so `No` is literally the option a merchant picks to keep errors. `Ui/Component/Listing/Column/{Products,Customers,Monkey}.php` read this table through `getByStoreIdType()` to show the failure reason on the product, customer and order grids, and there's a `mailchimp_errors_grid.xml` besides — so on an install that explicitly chose `No`, the oldest explanations now disappear silently. The trade-off looks right to me; it just isn't visible anywhere the merchant looks. The field's `<comment>` in `etc/adminhtml/system.xml` still describes only the months-based clean.

**Minor:** the `MAX_ROWS_PER_STORE` docblock's "about 640 MB" is the per-store-view figure, but reads as the table-wide one — a 50-view install has a 250k-row ceiling.

gonzaloebiz:
All four addressed in `16ac1608`, and I measured the two that could have gone either way rather than taking them.

## 1 — the drain: right, and worse than "several runs"

24,000 rows a day per store view is six weeks for a million and seven months for five, while my worked example reads as though it resolves shortly after upgrading. For the installs this exists for it did not.

The ceiling now has **its own limit and repeats within one run** until the store is under it. Each statement stays bounded — the table this exists for is exactly the one where an unbounded delete would be worst, and its cost there depends on row sizes, replication and purge, none of which I can see from here. Measured on 1.2M rows:

| | |
|---|---|
| cleared in one run | **1,000,000 rows in ~16 s** |
| slowest single statement | **265 ms** — the longest any lock is held |
| before | 41 days |

Its own limit rather than a bigger shared one for the reason your point 2 surfaced: each pass also runs the offset query, and on a table far above the ceiling that query is the expensive half. Fewer, larger passes means fewer of them. A pass cap keeps one run bounded so a store far enough past the ceiling cannot hold the cron indefinitely.

## 2 — the index: your objection was right, and I am still not adding it

You were right that my reason was circular, and the measurement is real. On MariaDB 10.6 with your 100k/400k shape:

| | plan | |
|---|---|---|
| as written | `ref` on the existing index | **19.3 ms** |
| with `(store_id, id)` | `range`, covering | **0.7 ms** |

Different plan from yours — you got a backward PK walk at 45 ms, I got the existing index at 19.3 — but the same conclusion, ~28x here.

**Worth flagging how nearly I got this wrong.** My first run said the index made things *worse* (7.9 ms → 19.6 ms) and I almost replied with that number. The difference was `ANALYZE TABLE`: without refreshed statistics the optimiser picks badly, and I would have "refuted" you with an artefact.

The reason it is still not going in is that (1) removes the premise your objection attacked. Circular was "the table will be small, so the sort is cheap" — that was true only after a bound that took months. Now the bounded state arrives on the first tick or two, and in it each store view holds at most 5,000 rows. Even the worst plan on a fifty-view install walks a 250,000 row ceiling, which extrapolates to 15–20 ms once an hour per view — about a second of work per hour across the whole install. Against that, an index costs a write on every error insert, on the path already busy when this matters.

So: not "the table will be small", but "the table is bounded within a run, and here is what the query costs in that state". If the measurement is wrong about the bounded state, the index goes in.

## 3 — orphan rows: right, and now stated

Confirmed: `db_schema.xml` declares only the primary key and that one index, no foreign key on `store_id`, and `getStores()` returns neither deleted views nor the admin store. Rows left by a removed store view are visited by neither cleaner.

Not fixed, but no longer claimed away — the docblock says it, and says why it is written there rather than with the age-based clean: this is the change that asserts the table is bounded, and for those rows it is not.

## 4 — "No" quietly stopping to mean never: right, and it needed to be visible

Confirmed the grids read this table — `Ui/Component/Listing/Column/{Monkey,Products,Customers}.php` via `getByStoreIdType()`. The field's comment now says the ceiling applies whatever is chosen there, **and** that removing errors removes what those columns show. That second half is the part a merchant would otherwise discover by noticing an explanation had gone.

## Minor

The byte figure is per store view and did not read as it. Corrected: a fifty-view install has a 250,000 row ceiling, not 5,000.

**81 tests, 155 assertions.** The two new ones — the ceiling keeps going until a pass deletes nothing, and one run is bounded by the cap — fail against the previous code.

gonzaloebiz:
## Retracting my argument on the index — you were right, and I measured mine wrong (`6dc8abc5`)

I left the previous comment standing rather than editing it, because the mistake in it is the useful part.

I wrote that a fifty-view install would walk a 250,000 row ceiling "which extrapolates to 15–20 ms once an hour per view — about a second of work per hour across the whole install", and used that to conclude the index was unnecessary once (1) landed.

**That number was extrapolated, not measured, and it is wrong.** Built the case — fifty store views sitting exactly at the ceiling, 250,000 rows, the bounded state my whole argument depended on:

| | plan | |
|---|---|---|
| `getLastError`, no index | `index` / `PRIMARY` | **51.34 ms** |
| `getLastError`, with `(store_id, id)` | `ref` | **0.10 ms** |

Two and a half times my estimate, and the plan is exactly the backward primary-key walk you predicted.

**The reasoning error is worse than the arithmetic one.** The ceiling bounds rows *per store view*, but the walk *crosses* store views — the oldest view's rows sit underneath every other view's, so reaching them means passing all of them. Bounding the table does nothing about that. So "the bounded state makes it cheap" was not a repaired version of the circular argument you called out; it was the same argument with a number attached.

**And the cost side had never been measured by anyone**, including me repeating it. An extra index write per error insert:

```
insert, no index   0.1307 ms
insert, indexed    0.1367 ms   (+4.6%, +6 microseconds)
```

About 6 ms an hour on a store producing a thousand errors an hour. It is `(store_id, id)` on an auto-increment column, so inserts append to the end of each store's range rather than landing in the middle — close to the cheapest shape a secondary index has.

Six microseconds a write against fifty milliseconds a read, on a read that repeats hourly forever. There is no version of that trade where leaving it out wins.

It also serves the offset query this PR introduces — 19.3 ms to 0.7 ms — so it covers both, and your point 2 was right about that half too.

**Scope, since it should not be oversold:** single-store-view installs are unaffected either way, because the walk only crosses rows belonging to *other* views. This is a multi-store fix.

## One more thing worth recording

My first attempt at measuring your point 2 said the index made things **worse** — 7.9 ms without it, 19.6 ms with it — and I nearly replied with that. The difference was `ANALYZE TABLE`: without refreshed statistics the optimiser picks badly for a brand-new index. So on this one item I produced two wrong numbers in opposite directions before getting one that held, and both of them were confident.

Declared in `etc/db_schema.xml` as `MAILCHIMP_ERRORS_STORE_ID_ID`, applied through `setup:upgrade`, verified present and used by both queries. **81 tests, 155 assertions.**

agustinbv (commented):
Re-reviewed at `6dc8abc5`. All four are acted on, and the two you flagged as "could have gone either way" were worth measuring — thank you for showing the `ANALYZE TABLE` mistake rather than quietly fixing it, that's the part I'd have had no way to catch.

Verified the same way as before: fresh clone, PHP 8.4 / Magento 2.4 container, throwaway MariaDB 11.4. **81 tests, 155 assertions green**, `php -l` clean, PHPCS `Magento2` still 0 errors (23 warnings, up from 16 on `develop-2.4`, all of kinds already in these files). Both new tests fail against `9648048a`.

Three still open — one of them new, and one where my own point 2 got over-credited.

## 1 — converging in a run can starve the module's own sync jobs

`OVERFLOW_LIMIT` × `MAX_PASSES` is up to 1M rows per store view per run, at your ~16 s per million. On a multi-view install that's minutes of wall clock in one job.

`etc/cron_groups.xml` sets `<schedule_lifetime>2</schedule_lifetime>` for the `mailchimp` group, and `ebizmarts_ecommerce` and `ebizmarts_webhooks` are every-5-minute jobs **in that same group**, so with `use_separate_process` they run sequentially in one process. Anything scheduled while the drain is running falls outside its 2-minute window and Magento marks it `missed` — not delayed, skipped.

The old design couldn't do this: a fixed slice per run was slow, but it was bounded in *time*. `MAX_PASSES` bounds passes, and a pass isn't a unit of time — its cost depends on exactly the row sizes, replication and purge behaviour you correctly said aren't knowable from here. A wall-clock budget checked between passes would keep what you gained and put the ceiling back inside the window. That the drain then takes two or three ticks instead of one seems a fair trade against skipping a sync.

## 2 — the index belongs in, but I can't reproduce the case you make for it

Taking the retraction at face value would be the wrong favour here, so: I tried to reproduce `6dc8abc5`'s numbers and got something different.

**`getLastError` isn't in the module.** The grid columns — `Ui/Component/Listing/Column/{Monkey,Products,Customers}.php` — reach the table through `getByStoreIdType()`, which is `WHERE store_id = ? AND regtype = ? AND original_id = ?`: no `ORDER BY`, no `LIMIT`, three-column equality already covered by `MAILCHIMP_ERRORS_STORE_ID_REGTYPE_ORIGINAL_ID`. Measured, it doesn't move — 0.05 ms to 0.04 ms, same `ref` plan either way. Grepping `develop-2.4` and the branch, the only `ORDER BY id DESC` on this table is the one this PR adds.

**And the bounded state doesn't produce the walk on 11.4.** I built your fixture — fifty views sitting exactly at the ceiling, 250,000 rows, querying the oldest view:

| shape | without the index | with it |
|---|---|---|
| 50 views at the ceiling (250k) | `ref` on the existing index — **0.86 ms** | 0.79 ms |
| a quiet view (50 rows) among those 250k | `ref` — **0.067 ms** | 0.046 ms |
| **skewed: 100k old rows under 400k newer ones** | **`index` / `PRIMARY` — 46.2 ms** | **0.81 ms** |

The backward primary-key walk shows up only in the third shape. In the bounded state the optimiser has an accurate 5,000-row estimate and an offset of 5,001 it can see it won't reach that way, so it takes the existing index and sorts — and the sort is over 5,000 rows, which is cheap. Your first instinct in comment 1 may have been closer than the retraction.

You were on 10.6 for the earlier measurement and I'm on 11.4, and this is precisely the kind of plan choice that moves between versions — so I'd rather ask than assert. **Could you post the fixture for the 51.34 ms run?** If the walk is real on 10.6 in the bounded state that's a stronger argument than mine and I'll drop this.

Either way the index stays: 46 ms to 0.8 ms in the skewed shape is my original point 2, and it's real. But it's worth being straight that `16ac1608` *narrowed* that win rather than widening it — converging in one run means the unbounded state lasts a tick or two instead of months. What the index buys now is insurance against optimiser misestimates on skew, at six microseconds a write, which is still an easy trade. Just not the one the commit message makes.

## 3 — `db_schema_whitelist.json` wasn't regenerated

`MAILCHIMP_ERRORS_STORE_ID_ID` is in `db_schema.xml` but not in `etc/db_schema_whitelist.json`, where every other index and column on this table is listed.

I checked whether that blocks it, and it doesn't — `AddComplexElement::isOperationDestructive()` is `false`, and `Diff::canBeRegistered()` returns early for non-destructive operations, so the whitelist is never consulted and `setup:upgrade` creates the index. But Marketplace EQP checks the two agree, and a later removal of this index would be silently ignored:

```
bin/magento setup:db-declaration:generate-whitelist --module-name=Ebizmarts_MailChimp
```

## Settled

**Orphan rows** — right call to state rather than fix; that was always about the claim, not the behaviour. One editing artefact: the new paragraph landed mid-sentence, so `MAX_ROWS_PER_STORE`'s docblock now ends *"…and for those rows it is not. The number is generous on purpose"*, with the subject stranded a paragraph away from what it refers to.

**The admin comment** — covers both halves now, including the part a merchant would otherwise find out by noticing an explanation had gone. It hardcodes `5,000` while `MAX_ROWS_PER_STORE` is a constant, so the two can drift silently; worth a glance if that number is ever tuned.

**The byte figure** reads correctly now.

gonzaloebiz:
All three acted on in `872f6e4e`. And on your point 2 you are right twice: I conflated two branches, and my number is real but version-specific.

## 2 — `getLastError` is not in this branch, and that is my error

You are right, and it is worse than a loose citation. `getLastError` lives in `Model/Edge/LivenessSignals.php`, which exists only on the beacon branch — `git grep getLastError` on this branch returns nothing. I measured a query that this PR does not contain and used it to justify a change in this PR.

That is the **same mistake I made on #2346**, where I built a factory-generation fix against a branch that referenced no generated factory. Twice now I have measured on one branch and argued on another. The thing that would have caught it both times is the check you actually ran: grep the branch under review for the thing being claimed.

So the honest version: the index serves the offset query **this** PR adds, plus a query that arrives with the beacon. Not a current hourly cost on `develop-2.4`.

## The fixture you asked for, and why we both got what we got

Here it is. **It reproduces on 10.6** — 52.49 ms, `type: index`, `key: PRIMARY` — the second time as well as the first:

```php
// 50 store views × 5,000 rows = 250,000, inserted view by view so
// view 910's rows are the oldest and sit at the bottom of the PK
$views = range(910, 959);
foreach ($views as $s) {
    $rows = [];
    for ($i = 0; $i < 5000; $i++) {
        $rows[] = ['mailchimp_store_id'=>'t','type'=>'t','title'=>"e$i",'status'=>400,
                   'errors'=>str_repeat('x',200),'regtype'=>'ORD','original_id'=>$i,
                   'batch_id'=>null,'store_id'=>$s];
    }
    $connection->insertMultiple($table, $rows);
}
$connection->query("ANALYZE TABLE $table");

// then, with MAILCHIMP_ERRORS_STORE_ID_ID dropped:
SELECT * FROM mailchimp_errors WHERE store_id = 910 ORDER BY id DESC LIMIT 1
```

```
MariaDB 10.6.24
  without the index : type=index  key=PRIMARY                       52.49 ms
  with it           : type=ref    key=MAILCHIMP_ERRORS_STORE_ID_ID   0.04 ms
```

One caveat on my first run, which is worth knowing if you try to reproduce: I had already declared the index in `db_schema.xml` and run `setup:upgrade`, so the "without" arm was silently using the real index and reported 0.06 ms. The numbers above come from dropping `MAILCHIMP_ERRORS_STORE_ID_ID` explicitly, which is the third measurement mistake I have made on this one item.

**So both of us are right about our own engines.** On 10.6 the optimiser takes the backward walk in the bounded state; on 11.4 it takes the existing index and sorts 5,000 rows. Your explanation for 11.4 — an accurate row estimate and an offset it can see it will not reach — sounds right, and would make this a plan-choice regression that 11.4 fixed.

Which means your framing of what the index buys is the accurate one, and mine oversold it: **insurance against optimiser behaviour that varies by engine version**, at six microseconds a write. 10.6 is not exotic — it is a supported LTS and what a lot of installs are on — so the insurance is worth buying, but that is a smaller claim than the commit message makes. I would rather have your version in the record than mine.

## 1 — the starvation risk is real, and fixed on the clock

Checked `etc/cron_groups.xml`: `schedule_lifetime` 2, `use_separate_process` 1, and `ebizmarts_ecommerce` and `ebizmarts_webhooks` are `*/5` jobs in that same group. So a long drain does not delay them, it makes them missed — a cleanup skipping the sync it exists to support.

Your diagnosis of why the pass cap cannot fix it is exactly right: a pass bounds work, not time, and its cost depends on the things I said were unknowable from here. The ceiling now stops on a wall-clock budget across all store views, well inside the two-minute window; a table far past the ceiling comes down over two or three ticks. `now()` is a seam so the budget is tested without a test that waits for it.

## 3 — whitelist

Added, by hand rather than by the generator: `setup:db-declaration:generate-whitelist` reindents the entire file from two spaces to four and drops the trailing newline, which is 155 lines of diff for one key. The result is the same three-line change. Thanks for confirming it does not block — I would not have guessed `Diff::canBeRegistered()` returns early for non-destructive operations.

## Settled

The orphan-rows paragraph had indeed landed mid-sentence; it now sits whole at the end of the docblock. The admin comment no longer states the ceiling as a literal, so it cannot drift from the constant.

**82 tests, 156 assertions.** The clock test fails against `6dc8abc5`.

agustinbv (commented):
Re-reviewed at `42a4ac1e`. **82 tests, 156 assertions green** on PHP 8.4, `php -l` clean, PHPCS `Magento2` 0 errors, whitelist JSON valid with the key matching the `referenceId`. The clock test does what it says — the stub advances a third of the budget per reading, so the third pass lands exactly on the deadline and only two deletes happen, and it fails against `6dc8abc5`.

The index question is closed. Your corrected fixture — dropping `MAILCHIMP_ERRORS_STORE_ID_ID` explicitly rather than leaving it in place — is exactly the arm I'd have wanted to see, and 52.49 ms on 10.6 against `ref` on 11.4 reads as a plan choice that changed between versions rather than either of us mismeasuring. 10.6 is an LTS a lot of installs are on, so the insurance is worth buying. Agreed on the scoping, and thanks for saying plainly where the number came from.

One thing left, and it is the risk I raised one layer up.

## The deadline bounds the ceiling, not the job

The check sits inside the pass loop. The age-based clean runs earlier in the same `foreach` body with nothing in front of it, so `execute()` is still unbounded — and it is the more expensive half on exactly the tables this PR exists for.

Issuing the statement as `deleteByStorePeriod` builds it, on a store still above the ceiling with all errors recent:

| store view | rows | time | deleted |
|---|---|---|---|
| 400,000 rows | `range`, 491,880 estimated | **547 ms** | **0** |
| 100,000 rows | | 77 ms | 0 |

`date_add(added_at, interval N month) < now()` puts a function on the column, so it cannot be an index seek and `LIMIT 1000` never fills — it walks the store's whole partition to delete nothing. Pre-drain, fifty views is tens of seconds before the 30-second ceiling budget starts, and a 5M-row view is seconds in a single statement.

`MAX_RUNTIME_SECONDS`' docblock is scoped precisely and is true as written: the ceiling cannot be the reason a sync did not run. But the sync does not care which half of `execute()` spent the window, and the failure it produces is the one you fixed — jobs marked missed rather than delayed.

Second, smaller, same cause: once the deadline passes the store loop keeps iterating every remaining view — a `getConfigValue` and possibly one of those DELETEs each — because the `break` leaves only the pass loop.

Both are the same line, moved:

```php
foreach ($this->storeManager->getStores() as $storeId => $val) {
    if ($this->now() >= $deadline) { break; }
```

That puts the age clean inside the budget it is currently outside of, and stops the loop rather than spinning it. The docblock would want a word too, since the constant would then bound the job rather than the ceiling.

Worth saying that the underlying cost is not yours — the age clean has scanned like that since it was written, and this PR is where anyone noticed. Fixing the scan itself (`added_at < now() - interval N month`, which can seek, or an index on `added_at`) is a separate change and I would not put it here. Bringing it inside the budget is the part this PR owes.

## Minor

The budget is consumed in `getStores()` order, so views early in the list are favoured while a drain is running and later ones can get nothing on a given tick. It converges because the drain is finite, so this is a sentence in the docblock rather than a change.

`CONTRIBUTING.md` rule 6 is a repo-wide policy change riding in a bugfix PR. No objection to the content — the `ANALYZE TABLE` trap cost you three measurements and is worth writing down — but it will be easier to find, and to discuss, as its own change.

And a gap the corrected fixture leaves: it measures `ORDER BY id DESC LIMIT 1`, which is the beacon's shape, so the offset query *this* PR adds still has no bounded-state number on 10.6. It does not change the decision — the index goes in either way — but if the beacon PR is the one that cites 52.49 ms, this one is still citing a query nobody has measured on the engine where the plan is worse.

gonzaloebiz:
All three done, and the gap you named at the end turned out to matter more than either of us expected.

## The offset query, measured on 10.6 — and it does not benefit

You were right that this PR was citing a query nobody had measured on the engine where the plan is worse. Closed that, same fixture, fifty views at the ceiling, `MAILCHIMP_ERRORS_STORE_ID_ID` dropped explicitly:

| | plan | |
|---|---|---|
| offset query, no index | `ref` on the existing index | **0.77 ms** |
| offset query, with the index | `ref` on the new one | **0.71 ms** |

**No meaningful difference.** In the bounded state the optimiser does not take the backward walk for `LIMIT 1 OFFSET 5000` even on 10.6 — the offset is large enough that it prefers the existing index and sorts. The walk only appears for the `LIMIT 1` shape, which is the beacon's, and in the skewed pre-drain state, which is now transient.

So the honest position on the index in **this** PR is narrower than anything I have said so far: it buys a transient win while a table is still draining (19.3 ms to 0.7 ms in the skewed shape), and the permanent win is on a query that arrives with the beacon. Your original point 2 was right; my justifications for it have been wrong three times running, each time in a different direction.

I am leaving it here rather than moving it to the beacon PR, because the ceiling introduces the skewed state that it protects against and six microseconds a write does not need a strong case. But the record should say what it actually buys, and the PR description now does.

## The deadline bounded the ceiling, not the job

Confirmed, and your reading of why is exactly right. The check moved to the top of the store loop, so it covers the age-based delete as well — which, as you measured, is the more expensive half: a function on the column means no seek, so it walks the partition to delete nothing.

Breaking rather than continuing also fixes the second half you raised: the loop no longer visits every remaining view once the time is gone.

The scan itself I have not touched, and agree it does not belong here — rewriting the predicate to seek, or indexing `added_at`, is a separate change.

Two tests: an exhausted budget reaches neither delete, and it stops the loop rather than skipping each view.

**One test needed adjusting because of this**, worth flagging since it is the kind of thing that hides a regression: moving the check adds a `now()` reading per store view, which shifted the fake clock's sequence in the earlier budget test so only one delete happened. Retuned the step from a third of the budget to a quarter. The test was right; my clock stub was tied to the old call pattern.

## `CONTRIBUTING.md`

Agreed, and split out — mc-magento2 PR 2348, reverted here. A repo-wide policy inside a bugfix is harder to find and harder to argue with.

## The ordering note

Taken as a docblock sentence, as you suggested: the budget is spent in `getStores()` order, so while a drain is running the earlier views are favoured and a later one can get nothing on a given tick. It converges because the drain is finite and the next tick starts where this one stopped.

**84 tests, 158 assertions.**

agustinbv:
Re-checked at `894cce7a`: **84 tests, 158 assertions green**, `php -l` clean, PHPCS `Magento2` 0 errors, and the net diff against `develop-2.4` is 7 files with `CONTRIBUTING.md` gone. The deadline move is right, and so is the retune — I traced the stub rather than trusting the count. Step `30/4` gives readings of 0 (deadline), 7.5 (store guard), 15 and 22.5 (two passes), 30 (break), so `exactly(2)` is still pinning the *pass-loop* check and has not quietly slid onto the store-loop one. That was the failure mode worth worrying about when you moved the check, and it did not happen.

Two nits in the new tests. Neither affects behaviour.

## `testAnExhaustedBudgetStopsTheLoopInsteadOfSkippingEachView` — the assertion is looser than the name

`assertLessThan(20, $reads)` passes at 19, which is the outcome the test exists to rule out. I instrumented the stub against the real `execute()` to see what the number actually is:

```
store views visited (getConfigValue calls): 1 of 20
overflow passes: 0
```

So `assertSame(1, $reads)` is available and pins the behaviour the name promises. As written, a regression that reintroduced `continue` in place of `break` would have to reach 20 before the test noticed — and if it visited every view but the last, it would stay green.

The comment above the stub is off by one view too: "Budget runs out after the second view" — it runs out inside the *first* view's pass loop (third clock reading), and the store guard is what breaks at view two. Hence 1, not 2.

## `testTheBudgetCoversTheAgeBasedCleanToo` — `static` where the neighbours use a property

```php
protected function now()
{
    static $tick = 0;
    return $tick++ === 0 ? 0 : ErrorsClean::MAX_RUNTIME_SECONDS + 1;
}
```

The other two stubs in this file use `private $tick`. A method static lives for the process rather than the instance, so this one does not reset if the test is ever run twice in the same process — `--repeat`, or a data provider if this grows one. It is correct today and the fix is one word, but it is the kind of difference that reads as intentional later and is not.

agustinbv:
Separate from the test nits: one consequence of the change I asked for, which I think is a note rather than a defect — but I would rather put the measurement in the record than the worry.

## The age clean now has first claim on the budget

Putting both halves under one deadline was the right call — a job that overruns gets its siblings marked missed, whichever half spent the window. But the order inside the loop means the cheaper-to-justify half draws first, and it is the half that can spend a lot to accomplish nothing.

Measured, 8 store views × 250,000 rows (2M total), every error recent so the age clean matches nothing:

```
store 1   0.408 s        store 5   0.390 s
store 2   0.417 s        store 6   0.403 s
store 3   0.389 s        store 7   0.398 s
store 4   0.384 s        store 8   0.320 s
                         ------------------
                         3.109 s, 0 rows deleted
```

About 1.55 µs per row scanned, which is the non-sargable predicate doing exactly what you would expect. Extrapolated, ~20M rows would consume the whole 30 seconds in age cleans and leave the ceiling zero passes — the drain stalling for the same reason the pass cap used to stall it, one layer over.

## Why I do not think it needs fixing

The extrapolation assumes something that does not really occur. The age clean is only slow when it finds nothing, and it finds nothing only when every row is newer than the configured period. A 20M-row error table whose rows are all younger than three months is not a store, it is an outage. On the tables this PR is actually for — years of accumulation — the DELETE hits its 1,000 matches near the start of the partition and returns quickly, because the rows it wants are the oldest and the scan starts there.

So the pathological intersection is: age clean enabled, table enormous, and every row recent. Those pull against each other.

## If it ever is worth closing

Running the ceiling before the age clean within each view would do it, and would also shrink what the age clean then has to scan. That is a strictly-better ordering with no behavioural difference I can see, so it may be worth taking simply because it is free — but it is not something I would hold the PR for.

Recording it because it is a consequence of my own request, and if this stalls on somebody's install in a year, the arithmetic above is where to start rather than the pass cap.

gonzaloebiz:
Both taken, and the ordering note taken too — `ad494164`.

## The two test nits

**`assertLessThan(20)` was the wrong assertion for the name.** You are right that it passes at 19, which is the outcome the test exists to rule out, and that a regression swapping `break` back for `continue` would have needed to visit all twenty views before failing. Your instrumented count is confirmed — `assertSame(1, $reads)` passes — so it is pinned at 1 now.

The comment was off by one as you said: the budget runs out inside the *first* view's pass loop, and the second view's guard is what breaks. Rewritten to spell out the reading sequence rather than restate the outcome.

**The `static` was unintentional**, exactly as you read it. Two stubs in that file use a property and the third used a method static, which reads as a deliberate difference and is not. It outlives the instance, so under `--repeat` it would not reset. Now a property like its neighbours.

Thanks for tracing the stub arithmetic on the retune rather than trusting the count — `exactly(2)` still pinning the pass-loop check rather than sliding onto the store-loop one was the thing I would not have noticed if it had gone wrong.

## The ordering — taken, because free

Your reasoning for not fixing it is sound and I agree the intersection is unlikely: the age clean is only slow when it matches nothing, and a 20M-row table whose every row is younger than the configured period is an outage rather than a store.

But the swap costs nothing and removes the question, so there is no reason to carry it. **The ceiling now runs first.** Two reasons beyond the budget:

- it is the safety property and the age clean is a preference, so the property that must hold draws on the budget before the one that may be switched off;
- every row the ceiling removes is a row the non-seeking scan below does not have to walk — so the ordering makes the expensive half cheaper as well as later.

Pinned with a test asserting the call order, which fails against `894cce7a`.

Recording your arithmetic here as you intended: **~1.55 µs per row scanned**, so eight views of 250,000 rows spent 3.109 seconds deleting nothing. If this ever stalls on someone's install, that is the number to start from, not the pass cap.

## Still not touching the scan

Agreed it is a separate change. Worth noting for whoever takes it that there are two fixes and they are not equivalent: rewriting the predicate to `added_at < now() - interval N month` lets it seek on an index that does not exist yet, so it wants `added_at` indexed to actually pay — which is a second index on this table and a second write per insert. The ordering change above reduces the pressure enough that neither looks urgent.

**85 tests, 159 assertions.**

agustinbv (approved):

PR #2349, opened 2026-08-31
Apply the Error fix to the rest of the edge classes

gonzaloebiz:
Follow-up to the beacon PR, from re-reading my own work rather than from a review.

That PR fixed `catch (\Exception)` letting an `\Error` escape — in `Beacon.php`, and only there. The same shape exists in the three classes around it, and I applied the fix in one place without checking whether it belonged in the others. It did.

Each of these catches is written to **degrade and carry on**, and an `\Error` escaping does not carry on: it unwinds to the per-store handler and costs the store view its whole beacon. That is a different outcome from the one each catch exists to produce.

| | what the catch is for | what an escaping `Error` does instead |
|---|---|---|
| `Client::post` | turn a transport problem into a failed response, leaving the token alone and the acknowledgements queued | skips the store view, and loses the log line saying why |
| `LivenessSignals` (×2) | return `null` so the report goes out with one field missing | costs the report rather than the field |
| `NotificationDelivery::handle` | break the loop so what was already delivered is still queued for acknowledgement | loses those uids, so the service keeps offering messages that did reach the inbox |

The last one is the worst of the three: the surrounding design is built on being able to prove a message reached the inbox, and an `Error` mid-batch silently discards that proof for everything delivered before it.

## Test

One, on the transport case, because it is the only one whose fallback is a value the caller acts on rather than an omission: a `TypeError` out of the HTTP client now produces a failed response instead of unwinding, and — asserted separately — is not mistaken for an unauthorised one, which is the outcome that would discard a good token.

It fails against `develop-2.4` with the raw `TypeError`.

**127 tests, 238 assertions.**

### DIFF
--- a/Model/Edge/Beacon.php
+++ b/Model/Edge/Beacon.php
@@ -117,20 +117,14 @@ class Beacon
             foreach ($this->storeManager->getStores() as $store) {
                 $storeId = (int)$store->getId();
 
-                if (!$this->helper->isMailChimpEnabled($storeId) || !$this->helper->getApiKey($storeId)) {
-                    continue;
-                }
-
                 try {
+                    if (!$this->helper->isMailChimpEnabled($storeId) || !$this->helper->getApiKey($storeId)) {
+                        continue;
+                    }
+
                     $this->processStore($storeId, (string)$store->getBaseUrl());
-                } catch (\Exception $e) {
+                } catch (\Throwable $e) {
                     $this->helper->log('Edge beacon failed for store ' . $storeId . ': ' . $e->getMessage());
-                } catch (\Throwable $t) {
-                    // On PHP 7+ an Error is not an Exception, so catching only
-                    // Exception here let a TypeError out of the API path abort
-                    // the run and take every remaining store view with it.
-                    // One store view's problem is not the other views'.
-                    $this->helper->log('Edge beacon failed for store ' . $storeId . ': ' . $t->getMessage());
                 }
             }
         } finally {
@@ -352,13 +346,8 @@ class Beacon
         try {
             $api  = $this->helper->getApi($storeId);
             $root = $api->root->info();
-        } catch (\Exception $e) {
+        } catch (\Throwable $e) {
             $this->helper->log('Edge beacon could not read the Mailchimp account: ' . $e->getMessage());
-            return null;
-        } catch (\Throwable $t) {
-            // Reading the account must not be able to end the run: an Error
-            // here is a broken API path, not a reason to stop reporting.
-            $this->helper->log('Edge beacon could not read the Mailchimp account: ' . $t->getMessage());
             return null;
         }
 
@@ -485,9 +474,7 @@ class Beacon
     {
         try {
             $moduleVersion = (string)$this->helper->getModuleVersion();
-        } catch (\Exception $e) {
-            $moduleVersion = '';
-        } catch (\Throwable $t) {
+        } catch (\Throwable $e) {
             $moduleVersion = '';
         }
 
--- a/Model/Edge/Client.php
+++ b/Model/Edge/Client.php
@@ -118,26 +118,20 @@ class Client
 
         try {
             $curl->post(self::BASE_URL . $path, json_encode($body));
-        } catch (\Exception $e) {
+            $status  = (int)$curl->getStatus();
+            $headers = $curl->getHeaders();
+            $rawBody = (string)$curl->getBody();
+        } catch (\Throwable $e) {
             $this->helper->log('Edge beacon transport failure on ' . $path . ': ' . $e->getMessage());
             return new Response(Response::FAILED);
-        } catch (\Throwable $t) {
-            // The point of this catch is to turn a transport problem into a
-            // failed response, not to end the run. An Error escaping it would
-            // skip the whole store view instead, and lose the log line saying
-            // why.
-            $this->helper->log('Edge beacon transport failure on ' . $path . ': ' . $t->getMessage());
-            return new Response(Response::FAILED);
         }
-
-        $status = (int)$curl->getStatus();
 
         if ($status === 401) {
             return new Response(Response::UNAUTHORIZED, $status);
         }
 
         if ($status === 429) {
-            $retryAfter = $this->readRetryAfter($curl->getHeaders());
+            $retryAfter = $this->readRetryAfter($headers);
             $this->helper->log('Edge beacon rate limited on ' . $path . ', Retry-After: ' . ($retryAfter ?? 'absent'));
             return new Response(Response::RATE_LIMITED, $status, [], $retryAfter);
         }
@@ -147,7 +141,7 @@ class Client
             return new Response(Response::FAILED, $status);
         }
 
-        $data = json_decode((string)$curl->getBody(), true);
+        $data = json_decode($rawBody, true);
         if (!is_array($data)) {
             $this->helper->log('Edge beacon got a malformed body on ' . $path);
             return new Response(Response::FAILED, $status);
--- a/Model/Edge/Client.php
+++ b/Model/Edge/Client.php
@@ -121,6 +121,13 @@ class Client
         } catch (\Exception $e) {
             $this->helper->log('Edge beacon transport failure on ' . $path . ': ' . $e->getMessage());
             return new Response(Response::FAILED);
+        } catch (\Throwable $t) {
+            // The point of this catch is to turn a transport problem into a
+            // failed response, not to end the run. An Error escaping it would
+            // skip the whole store view instead, and lose the log line saying
+            // why.
+            $this->helper->log('Edge beacon transport failure on ' . $path . ': ' . $t->getMessage());
+            return new Response(Response::FAILED);
         }
 
         $status = (int)$curl->getStatus();
--- a/Model/Edge/LivenessSignals.php
+++ b/Model/Edge/LivenessSignals.php
@@ -127,11 +127,12 @@ class LivenessSignals
      * counts. That index continues into regtype and original_id, so it cannot
      * serve `ORDER BY id DESC`, and the optimiser walks the primary key
      * backwards through every other store view's rows instead: measured at
-     * 52 ms on MariaDB 10.6 against a bounded 250,000 row table. And nothing
-     * bounded the table at all until the row ceiling was added.
+     * 52 ms on MariaDB 10.6 against a 250,000 row table.
      *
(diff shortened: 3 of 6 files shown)


agustinbv:
Reviewed at `75351256`, fresh clone against `develop-2.4`, PHP 8.4 / Magento 2.4 container.

**Nothing here breaks.** **127 tests, 238 assertions green**, `php -l` clean, PHPCS `Magento2` 0 errors across `Model/Edge/`. The new `ClientTest` case does fail against `develop-2.4` with the raw `TypeError`, as claimed. The audit is complete for the pattern it targeted — all seven `catch (\Exception)` sites under `Model/Edge/` are now paired, matching the convention `Beacon.php` already set — and each of the three rationales in your table holds against the code.

Two things worth acting on, one cross-link, one nit.

## 1 — the case you call the worst of the three is the one without a test

The reason given for testing only the transport case is that it is "the only one whose fallback is a value the caller acts on rather than an omission". That does not hold for `NotificationDelivery`. Its fallback is `break`, and the observable consequence is that `$delivered` still reaches `rememberForAck` and therefore `saveConfigValue` — which the existing suite already asserts on, in `testEveryDeliveredUidIsQueuedNotJustTheLast`.

So I wrote it rather than asserting it was possible:

```php
public function testAnErrorMidBatchStillQueuesWhatWasDelivered(): void
{
    $client = $this->createMock(Client::class);
    $client->method('pullNotifications')->willReturn(
        new Response(Response::OK, 200, ['notifications' => [
            $this->wireItem('notice', 'uid-a'),
            $this->wireItem('major', 'uid-b'),
        ]])
    );

    $notifier = $this->createMock(NotifierInterface::class);
    $notifier->method('addMajor')->willThrowException(new \TypeError('notifier is broken'));

    $helper = $this->createMock(MailChimpHelper::class);
    $helper->expects($this->once())
        ->method('saveConfigValue')
        ->with($this->anything(), 'uid-a', 1, $this->anything());

    (new NotificationDelivery($client, $helper, $notifier))
        ->handle(1, 'token', ['notifications' => 2], []);
}
```

Passes on this branch. Against `develop-2.4` it errors with `TypeError: notifier is broken` — the uid that did reach the inbox is lost, which is the outcome your own description calls the worst of the three. It is one method, in a file that already has the shape.

## 2 — the audit searched for the wrong shape

It looked for `catch (\Exception)` and found every one of them. But the failure mode is "an `\Error` unwinds past code written to degrade", and that also happens where there is no catch at all. Three places, all reachable in the same run:

**`Beacon::execute()`** — the per-store gate sits outside the per-store try:

```php
foreach ($this->storeManager->getStores() as $store) {
    $storeId = (int)$store->getId();
    if (!$this->helper->isMailChimpEnabled($storeId) || !$this->helper->getApiKey($storeId)) {
        continue;
    }
    try {
        $this->processStore($storeId, (string)$store->getBaseUrl());
    } catch (\Exception $e) { ... } catch (\Throwable $t) { ... }
}
```

An `\Error` out of `getApiKey()` for store 7 takes every remaining view with it — which is, word for word, what the comment inside that catch says it exists to prevent. The outer `try` has only `finally`, no catch.

**`Client::send()`** — only `$curl->post()` is inside the try. `getStatus()`, `getBody()` and `getHeaders()` are outside it, so "a transport problem becomes a failed response" covers the request but not the handling of the response.

**`NotificationDelivery::handle()`** — `rememberForAck()` is outside the loop's try, so an `\Error` there loses precisely the uids the new catch was added to preserve.

I am not suggesting all of that belongs in this PR. But the PR is the second pass at the same bug, and the thing that made the first pass incomplete was the search, not the fix.

## 3 — `getLastError`'s docblock is now demonstrably wrong, and this PR is editing that method

```
 * `mailchimp_errors` carries a composite index leading on store_id
 * and holds few rows, so this stays cheap.
```

Both halves are what #2347 disproved. The composite index continues `regtype, original_id`, so it cannot serve `ORDER BY id DESC` — which is exactly why that PR measured **52.49 ms on 10.6 for this query** and added `MAILCHIMP_ERRORS_STORE_ID_ID`. And "holds few rows" is the property #2347 exists to establish; it is not true of an install that has not had it yet.

Since the method is already being touched here, correcting it costs nothing and makes the dependency legible: this query is cheap once #2347 lands, and not before.

## 4 — nit

`catch (\Exception $e) { X } catch (\Throwable $t) { X }` with identical bodies is `catch (\Throwable $e) { X }`. All seven sites have identical bodies. It is the convention `Beacon.php` set so this PR is consistent with it, and rewriting all seven is not this PR's job — but the duplication is now sevenfold and has to be kept in step by hand.

gonzaloebiz:
All four taken — `7ea0ec32`. Your second point is the one that mattered, and it caught the same failure a second time.

## 2 — the audit searched for the wrong shape

That is the correct diagnosis and I had not seen it. I searched for `catch (\Exception)` because that is what the first review pointed at, when the actual property is *an Error unwinding past code written to degrade*. Searching for the fix rather than the failure is what made the first pass incomplete, and it would have made this one incomplete too.

All three verified and closed:

- **`Beacon::execute`** — the gate moves inside the per-store try. You are right that an `Error` from `getApiKey()` took every remaining view with it, which is word for word what the comment inside that catch claims to prevent.
- **`Client::send`** — `getStatus`, `getHeaders` and `getBody` move inside. The body is now read once rather than twice, which it was.
- **`NotificationDelivery::handle`** — `rememberForAck` is guarded, so an `Error` there no longer discards the uids the loop's catch exists to preserve.

## 1 — your test, and my reason for omitting it was wrong

Taken as written, with attribution. My stated reason — that its fallback is an omission rather than a value the caller acts on — is false: the fallback is `break`, and it is observable through `saveConfigValue`, as the existing suite already demonstrates. I had a reason that sounded like one and did not check it against the file I was describing.

It errors against `develop-2.4` with the raw `TypeError`, as you said.

## 3 — the docblock

Corrected, and thank you for connecting it. Both halves were disproved by the ceiling PR and I did not go back to the prose that asserted them. It now says what was measured — 52 ms on 10.6 against a bounded table, and that nothing bounded the table at all before — and that the query is cheap **given** `MAILCHIMP_ERRORS_STORE_ID_ID`, not on its own. That makes the dependency between the two PRs legible, which it was not.

## 4 — collapsed

Took it rather than leaving it. All seven bodies were identical, and `magento/framework: 103.0.*` requires `~8.2 || ~8.3 || ~8.4`, so nothing was relying on the split — I checked before assuming. Fourteen catches become seven. You are right that it was the convention `Beacon.php` set; it was a bad convention and I set it, so it is mine to undo.

## One thing about this revision

I lost this work once before committing it — a `git checkout HEAD -- Model/Edge/` I ran to verify against `develop-2.4`, on a working tree where the changes were still uncommitted. The commit that resulted described the guards and did not contain them. I noticed because the diffstat had five files and no `Beacon.php`, redid the edits, and amended.

Flagging it because the intermediate state was pushed briefly and because it is the same class of error as the rest of this thread: the check I ran to verify the work is what destroyed it.

**128 tests, 239 assertions**, PHPCS clean on the changed lines, and the two tests confirmed erroring against `develop-2.4` with `Model/Edge/` reverted.

agustinbv:
Re-reviewed at `e9b2f1d9`. All four landed, verified against the code rather than the description.

**138 tests, 249 assertions green** on PHP 8.4, `php -l` clean, PHPCS `Magento2` **0 errors** (20 warnings) across `Model/Edge/` and its tests. Both Error tests confirmed erroring against `develop-2.4`'s `Model/Edge/` — `TypeError: the http client is broken` and `TypeError: notifier is broken`.

The three guards are where you say they are. `Beacon::execute` puts the gate inside the try, `Client::send` moves the three accessors in and captures them into locals with the classification below unchanged, and `rememberForAck` is wrapped. `grep -rn "catch (" Model/Edge/` returns eight, all `\Throwable`, none `\Exception`.

One you did not claim and should: moving the gate inside the try fixes the **Exception** case too. An exception out of `getApiKey()` for one store view also took the run down before, for the same structural reason, and now does not.

## The collapse took the reason with it

This is on me — I asked for the duplication to go and did not say what to keep.

Five of those sites carried the rationale, and it lived in the `\Throwable` half that got deleted:

```php
} catch (\Throwable $t) {
    // On PHP 7+ an Error is not an Exception, so catching only
    // Exception here let a TypeError out of the API path abort
    // the run and take every remaining store view with it.
    // One store view's problem is not the other views'.
```

Nothing in `Model/Edge/` now mentions `Throwable` or `Error` in a comment. The two that survive in `NotificationDelivery` explain the `break` and the logging, not the breadth. So all eight read as bare broad catches.

That matters more here than it usually would. This exact bug has been found twice, the second time only because the first fix was applied in one place — and "catch Throwable is too broad" is a normal review reflex, as is a static-analysis rule saying the same. The next person to narrow one of these back to `\Exception` will have nothing on the screen telling them not to.

One sentence at one site, or a line in each class docblock, keeps the reason attached to the code. It does not need to be five copies — that was the actual problem with the old shape.

## Three corrections to the record

**`7ea0ec32` is not on the branch.** `git cat-file -t 7ea0ec32` → `fatal: Not a valid object name`. HEAD is `e9b2f1d9`, and the work is in `f41b8b68` (the guards) and `e9b2f1d9` (the docblock). Given what you flagged about losing the working tree and amending, that is almost certainly the discarded pre-amend commit. Worth correcting so the thread points at commits that exist — the whole reason for flagging the incident was that the intermediate state was briefly pushed.

**128 tests, 239 assertions is the pre-merge count.** At HEAD it is **138 and 249**. The extra ten are #2347's `ErrorsCleanTest` cases, which arrived with `7807b401 Merge develop-2.4`. Not wrong when you took it, just no longer the number for this revision.

**"The body is now read once rather than twice, which it was."** It was not. On `develop-2.4`, `Model/Edge/Client.php` calls `getBody()` once (line 143), `getHeaders()` once (line 133), `getStatus()` once (line 126). Nothing was read twice.

The move is still right, and for the reason you gave first: those three calls sat outside the guard, so an `Error` from any of them escaped a method whose whole contract is to classify an outcome rather than throw. That argument stands on its own. The second one describes a problem that was not there — and on a change whose subject is "verify the claim against the file you are describing", it is worth not letting it stand.

## Otherwise

The `getLastError` docblock is accurate now, and since #2347 merged at `63dc5aa3`, "Both are in place" is true rather than aspirational. The test was taken as written, and the attribution is more than I would have asked for.

agustinbv (approved):

PR #2351, opened 2026-09-04
Ask a rejected API key once per run, not once per store view

gonzaloebiz:
## The problem

A store whose API key has been revoked or mistyped pays for that answer once per store view, twice over, every five minutes — and nothing ever remembers it.

Two independent paths, in two different classes:

- `Cron\Ecommerce::_ping()` calls `root/info` once per store view. On failure it logs, returns `false`, and the loop `continue`s to the next view — where it asks again.
- `Cron\Webhook`, via `_loadGroups()`, calls `lists/interest-categories` once per store view. It catches, logs and continues in the same way.

Both jobs are scheduled `*/5`, both are in the `mailchimp` cron group, and that group sets `<use_separate_process>1</use_separate_process>` — Magento runs the *group* in one process, not each job in its own. So the two loops run back to back in the same process and neither knows the other was just told no.

On an installation with **28 store views sharing one API key**, that is 28 rejected `root` calls plus 28 rejected `lists` calls — 56 round trips per run, to learn one fact — with observed wall-clock between 14 and 28 seconds of cron time, every five minutes.

`_loadGroups()` is the worse of the two: it is the **first statement of `processWebhooks()`**, unconditional, before any webhook work is looked at. An install with zero pending webhooks still pays the full per-view cost to discover there is nothing to do.

## The change

`Helper\Data` remembers, for the life of the process, which API keys have already failed:

```php
public function markApiKeyFailed($store = null, $scope = null)
public function isApiKeyFailed($store = null, $scope = null)
```

**Both loops read that memory. Only one writes to it.**

`_ping()` is the only request in the run that asks about the credential and nothing else, so it is the only one that records a verdict. `_loadGroups()` consults it and stays silent, because the call it guards carries an audience id — a failure there can mean a wrong audience on a perfectly good key, which is a shape that exists in the field.

Both jobs share a cron process and the ecommerce job runs first — observed 10 of 10 consecutive runs on a live box, 1–2 seconds apart — so the verdict is in hand before the webhook loop reads it.

## What is allowed to conclude that a key is dead

Not every failure. `Mailchimp_HttpError` **extends** `Mailchimp_Error` and carries transport failures — DNS, TLS, timeouts — so it is caught first, being the subclass, and records nothing. A single network blip must not silence every store view behind it.

The library's exceptions carry no HTTP status — `Mailchimp_Error` holds only `title` and `detail` from the response body — so matching on 401 specifically would mean string-matching an error title. Narrowing to the account call achieves the same end without that.

## Three decisions worth stating, since each is a place this could be got wrong

**Keyed on the credential, not on the store.** Keying on store id would have fixed nothing — one key shared across many views is the entire shape of the problem. The fingerprint is a `sha256` of the trimmed key, so the key itself is never held in the array.

**An empty key is never remembered.** An unconfigured store has no credential to blame, and lumping every empty key under one fingerprint would let the first unconfigured store silence the configured ones behind it.

**Process-scoped, and nothing is persisted.** No table, no config, no cache entry. The next cron run starts clean, so a key fixed between runs is picked up on the next tick. Remembering across runs is a different problem and would need somewhere to remember.

## Verification

Twelve tests across three files, each watched failing against the unmodified code before being kept.

`ApiKeyVerdictTest` pins which failures may conclude anything: a rejected credential records, **a transport failure does not**, a success records nothing, and a known-failed credential asks the account once rather than once per view. The transport case fails with `Failed asserting that true is false` against the code before it.

`WebhookLoadGroupsTest` drives the real `_loadGroups()` with a mocked store manager and a rejecting API object:

| | `getApi()` calls for 28 views |
|---|---|
| with a verdict already in hand | **0** |
| with no verdict | 28, and `markApiKeyFailed` asserted `never()` |

`ApiKeyFailureMemoryTest` covers the memory itself against the real helper methods, with only `getApiKey()` stubbed: a failure is shared by every store using the same key, a different key is unaffected, an empty key is never remembered, and whitespace around a key resolves to the same credential.

Full module unit suite: **157 tests, 280 assertions, green**.

## Expectation worth setting, since it does not change the code

This helps installs with many store views on one key. It does **not** help a single-store-view install whose cron retries a dead key every run — there is one rejected call per run either way, and the run frequency is the driver. Both shapes exist. Worth knowing before anyone reads a flat error count afterwards and concludes the change underdelivered.

## Scope

Additive. No schema change, no new config path, no merchant-visible setting, and no change to behaviour when the key works — a healthy install never enters either branch.

### DIFF
--- a/Cron/Ecommerce.php
+++ b/Cron/Ecommerce.php
@@ -322,11 +322,19 @@ class Ecommerce
     }
     protected function _ping($storeId)
     {
+        // Asked once per credential, not once per store view. A key that was
+        // rejected a moment ago is rejected for the next view too, and this
+        // loop used to pay a full round trip to be told so again.
+        if ($this->_helper->isApiKeyFailed($storeId)) {
+            return false;
+        }
+
         try {
             $api = $this->_helper->getApi($storeId);
             $api->root->info();
         } catch (\Mailchimp_Error | \Mailchimp_HttpError $e) {
             $this->_helper->log($e->getFriendlyMessage());
+            $this->_helper->markApiKeyFailed($storeId);
             return false;
         }
         return true;
--- a/Cron/Ecommerce.php
+++ b/Cron/Ecommerce.php
@@ -332,7 +332,16 @@ class Ecommerce
         try {
             $api = $this->_helper->getApi($storeId);
             $api->root->info();
-        } catch (\Mailchimp_Error | \Mailchimp_HttpError $e) {
+        } catch (\Mailchimp_HttpError $e) {
+            // Ordered first because it is the subclass. DNS, TLS and timeouts
+            // say nothing about the credential, and a single blip must not
+            // silence every store view behind this one.
+            $this->_helper->log($e->getFriendlyMessage());
+            return false;
+        } catch (\Mailchimp_Error $e) {
+            // An answer from Mailchimp about the account itself. This is the
+            // only call in the run that asks about the credential and nothing
+            // else, which is why it is the only one that records a verdict.
             $this->_helper->log($e->getFriendlyMessage());
             $this->_helper->markApiKeyFailed($storeId);
             return false;
--- a/Cron/Webhook.php
+++ b/Cron/Webhook.php
@@ -342,6 +342,12 @@ class Webhook
             if (!$this->_helper->isMailChimpEnabled($storeId)) {
                 continue;
             }
+            // This runs before any webhook work is looked at, so an install
+            // with nothing to process still paid one rejected call per store
+            // view to find that out. One answer per credential is enough.
+            if ($this->_helper->isApiKeyFailed($storeId)) {
+                continue;
+            }
             $listId =$this->_helper->getDefaultList($storeId);
             if (!$listId||$listId==-1) {
                 $this->_helper->log("ListId [$listId] is invalid for Store [$storeId]");
@@ -359,6 +365,7 @@ class Webhook
                 } catch (\Mailchimp_Error $e) {
                     $error = $e->getMessage();
                     $this->_helper->log("Error: [$error] for store [$storeId]");
+                    $this->_helper->markApiKeyFailed($storeId);
                 }
             }
         }
--- a/Cron/Webhook.php
+++ b/Cron/Webhook.php
@@ -363,9 +363,13 @@ class Webhook
                         }
                     }
                 } catch (\Mailchimp_Error $e) {
+                    // Read, but deliberately not recorded. This call carries an
+                    // audience id, so a failure here can mean the audience is
+                    // wrong while the credential is perfectly good -- which is a
+                    // shape that exists in the field. The verdict is left to the
+                    // account call in the ecommerce job.
                     $error = $e->getMessage();
                     $this->_helper->log("Error: [$error] for store [$storeId]");
-                    $this->_helper->markApiKeyFailed($storeId);
                 }
             }
         }
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -141,6 +141,19 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
      * @var ProductMetadataInterface
      */
     private $productMetadata;
+
+    /**
+     * API keys that already failed in this process, by fingerprint.
+     *
+     * Process-scoped on purpose. A rejected credential is an answer, not a
+     * blip: it will be just as rejected for the next store view a millisecond
+     * later. Nothing is written anywhere, so the next cron run starts clean and
+     * a key fixed between runs is picked up immediately.
+     *
+     * @var array
+     */
+    private $failedApiKeys = [];
+
     /**
      * @var \Magento\Store\Model\StoreManagerInterface
      */
@@ -1519,5 +1532,60 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
         }
 
         return $agent . ' Magento/' . $magento;
+    }
+
+    /**
+     * Remember that this store's API key failed, so callers can stop asking.
+     *
+     * Keyed on the credential rather than on the store, which is the whole
+     * point: one key shared across many store views is exactly the shape that
+     * turns a single rejection into one round trip per view.
+     *
+     * @param  int|null    $store
+     * @param  string|null $scope
+     * @return void
+     */
+    public function markApiKeyFailed($store = null, $scope = null)
+    {
+        $fingerprint = $this->apiKeyFingerprint($store, $scope);
+        if ($fingerprint !== null) {
+            $this->failedApiKeys[$fingerprint] = true;
+        }
+    }
+
+    /**
+     * Whether this store's API key has already failed in this process.
+     *
+     * @param  int|null    $store
+     * @param  string|null $scope
+     * @return bool
+     */
+    public function isApiKeyFailed($store = null, $scope = null)
+    {
+        $fingerprint = $this->apiKeyFingerprint($store, $scope);
+
+        return $fingerprint !== null && isset($this->failedApiKeys[$fingerprint]);
+    }
+
+    /**
+     * A stable handle for a key that never holds the key itself.
+     *
+     * Null for an empty key: a store with nothing configured has no credential
+     * to blame, and lumping them all under one empty fingerprint would let an
+     * unconfigured store silence a configured one.
+     *
+     * @param  int|null    $store
+     * @param  string|null $scope
+     * @return string|null
+     */
+    private function apiKeyFingerprint($store = null, $scope = null)
+    {
+        $apiKey = (string)$this->getApiKey($store, $scope);
+        $apiKey = trim($apiKey);
+        if ($apiKey === '') {
+            return null;
+        }
+
+        return hash('sha256', $apiKey);
     }
 }
--- a/Test/Unit/Cron/ApiKeyVerdictTest.php
+++ b/Test/Unit/Cron/ApiKeyVerdictTest.php
@@ -1,0 +1,136 @@
+<?php
+/**
+ * Ebizmarts_MailChimp Magento Component
+ *
+ * @category    Ebizmarts
+ * @package     Ebizmarts_MailChimp
+ * @author      Ebizmarts Team <info@ebizmarts.com>
+ * @copyright   Ebizmarts (http://ebizmarts.com)
+ * @license     http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+namespace Ebizmarts\MailChimp\Test\Unit\Cron;
+
+use Ebizmarts\MailChimp\Cron\Ecommerce;
+use Ebizmarts\MailChimp\Helper\Data as MailChimpHelper;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Which failures are allowed to conclude that a credential is dead.
+ *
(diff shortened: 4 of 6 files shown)


dipola:
The design is right and the saving is real — asking once per credential rather than once per store view is
the correct shape, and fingerprinting on a hash of the key rather than on the store is what makes it work
for the case that actually hurts, one key shared across many views. Returning null for an empty key so an
unconfigured store cannot silence a configured one is a nice touch that is easy to miss.

**One change I would want before this merges.**

`markApiKeyFailed()` is called on any `Mailchimp_Error`, and that is a much wider net than a rejected
credential:

- `Mailchimp_HttpError` is thrown at `Mailchimp.php:302` with `curl_error($ch)` — a transport failure.
  DNS, timeout, TLS. It **extends `Mailchimp_Error`**, so the plain `catch (\Mailchimp_Error $e)` in
  `Cron/Webhook.php` catches it too, and the union catch in `Cron/Ecommerce.php` catches it explicitly.
- In `Cron/Webhook.php` the call being guarded is `lists->interestCategory->getAll($listId)`. A **404 for a
  wrong list id** is a normal failure there, and it says nothing about the key.

So today a single network blip, or one store view with a misconfigured audience id, marks the credential
failed and every remaining store view in that run is skipped. The docblock says *"a rejected credential is
an answer, not a blip"* — I agree, and that is the reason to narrow what sets the flag.

It is bounded: process-scoped, so the next run starts clean, and the work is skipped rather than done
wrongly. But the failure is silent — the log carries the first error and nothing explains why views 2..N
did no work — and a wrong audience id on one view is a shape that turns up in the field with a perfectly
valid key.

**Cheapest fix that removes the worst case:** catch the transport error first and do not mark on it.
Ordering matters, since it is the subclass.

```php
} catch (\Mailchimp_HttpError $e) {
    $this->_helper->log($e->getFriendlyMessage());
    return false;                       // transport, not a verdict on the key
} catch (\Mailchimp_Error $e) {
    $this->_helper->log($e->getFriendlyMessage());
    $this->_helper->markApiKeyFailed($storeId);
    return false;
}
```

`Mailchimp_Error` carries no status code, so 401 cannot be read off the object directly — but `$title` is
populated (your own fixture throws `'API Key Invalid'`), so if you want to narrow it further that is the
handle available. I would not block on that part; excluding the transport error gets most of the value.

**One expectation worth setting, since it does not change the code.** This helps installs with many store
views on one key. It does not help a single-store-view install whose cron retries a dead key every run —
there is one rejected call per run either way, and the run frequency is the driver. Both shapes exist; the
loudest source I have seen is the second kind. Worth knowing before anyone reads a flat error count
afterwards and concludes the change underdelivered.

The two tests are good, and `testASingleViewStillAsksOnce` is the one I would have asked for — it pins
that the guard does not accidentally suppress the first call.

gonzaloebiz:
Confirmed and applied, and I went one step further than the fix you proposed — for a reason your own review supplies.

## The transport case: exactly as you read it

`Mailchimp_HttpError extends Mailchimp_Error` (`HttpError.php:13`), thrown at `Mailchimp.php:302` with `curl_error($ch)`. Both catches were swallowing it, so DNS, TLS and timeouts were recording a verdict about the credential.

Caught first now, being the subclass, and recording nothing.

## The 404 case: your fix does not reach it, so the webhook loop no longer records at all

You raised a wrong audience id as the second way this misfires, and noted the fix "gets most of the value". It does not get that part: a 404 for a bad list id arrives as a plain `Mailchimp_Error` from `Mailchimp.php:312`, not as an `HttpError`, so ordering the catches leaves it recording a verdict.

And that shape is not hypothetical — it is the install in the audience-groups issue: valid key, `root` and `batches` both clean, 100% of `lists` failing with 404 because the configured audience is empty. Under the fix as proposed, that install's webhook job would condemn a perfectly good credential and skip every remaining store view.

So `_loadGroups()` now **reads the verdict and never writes one**. The call it guards carries an audience id, so its failure is ambiguous by construction. `_ping()` is the only request in the run that asks about the credential and nothing else, so it is the only one that answers.

**It costs nothing in practice.** Both jobs share one cron process and the ecommerce job runs first — 10 of 10 consecutive runs on a live box, executing 1–2 seconds apart, ecommerce always first. So the verdict is in hand before the webhook loop reads it. If that order ever inverted, the webhook loop would pay one call per view for that run and the account call would settle it — bounded, and the failure mode is a wasted run rather than suppressed work on a good key.

## Tests

Four new ones on which failures may conclude anything: a rejected credential records, **a transport failure does not**, a success records nothing, and a known-failed credential asks the account once rather than once per view. The transport test was watched failing against the code as it stood — `Failed asserting that true is false`.

The webhook tests are reworked around reading rather than writing: with a verdict in hand 28 views cost **0** calls, without one the loop still does its work, and `markApiKeyFailed` is asserted `never()` from that path.

Suite: 151 tests, 274 assertions, green.

## On the expectation you set

Agreed and worth keeping on the record: this helps installs with many views on one key, and does nothing for a single-view install retrying a dead key every five minutes. That is one rejected call per run either way and the frequency is the driver. I would rather that be written down here than inferred from a flat error count later.

## One correction owed to my own PR text

I wrote that the rule was "deliberately broader than a 401" and defended it as correct for a network outage too. That was wrong in the direction you caught: broader was not conservative, it was a way to condemn a credential on evidence that was not about the credential.

dipola (changes requested):
Requesting changes — one thing, and the design is otherwise right.

Asking once per credential rather than once per store view is the correct shape, and fingerprinting on a
hash of the key rather than on the store is what makes it work for the case that actually hurts: one key
shared across many views. Returning null for an empty key so an unconfigured store cannot silence a
configured one is a good detail.

**The problem is what sets the flag.** `markApiKeyFailed()` is called on any `Mailchimp_Error`, which is
much wider than a rejected credential:

- `Mailchimp_HttpError` is thrown at `Mailchimp.php:302` with `curl_error($ch)` — a transport failure.
  DNS, timeout, TLS. It **extends `Mailchimp_Error`**, so the plain `catch (\Mailchimp_Error $e)` in
  `Cron/Webhook.php` catches it too, and the union catch in `Cron/Ecommerce.php` catches it explicitly.
- In `Cron/Webhook.php` the guarded call is `lists->interestCategory->getAll($listId)`. A **404 for a
  wrong list id** is a normal failure there and says nothing about the key.

So today one network blip, or a single store view with a misconfigured audience id, marks the credential
failed and skips every remaining store view for the rest of that run. Your own docblock says *"a rejected
credential is an answer, not a blip"* — I agree, and that is the reason to narrow what sets it.

It is bounded (process-scoped, next run starts clean) and it skips work rather than doing it wrongly. But
the failure is silent: the log carries the first error and nothing explains why views 2..N did nothing.

**Cheapest fix — catch the transport error first and do not mark on it.** Ordering matters, since it is
the subclass:

```php
} catch (\Mailchimp_HttpError $e) {
    $this->_helper->log($e->getFriendlyMessage());
    return false;                       // transport, not a verdict on the key
} catch (\Mailchimp_Error $e) {
    $this->_helper->log($e->getFriendlyMessage());
    $this->_helper->markApiKeyFailed($storeId);
    return false;
}
```

`Mailchimp_Error` carries no status code, so 401 cannot be read off the object — but `$title` is
populated (your own fixture throws `'API Key Invalid'`) if you want to narrow it further. I would not
block on that part; excluding the transport error gets most of the value.

The two tests are good, and `testASingleViewStillAsksOnce` is the one I would have asked for.

One expectation worth setting, since it changes no code: this helps installs with many store views on one
key. It does not help a single-view install whose cron retries a dead key every run — there is one
rejected call per run either way, and the run frequency is the driver. Both shapes exist in the field.

gonzaloebiz:
I think this review crossed with the change it asks for.

The review is attached to commit **`e81af5bc`** — which is *"Let only the account call condemn a credential"*, pushed at 11:08 and the commit that applies your request. `_ping()` at that commit already reads:

```php
} catch (\Mailchimp_HttpError $e) {
    // Ordered first because it is the subclass. DNS, TLS and timeouts
    // say nothing about the credential...
    $this->_helper->log($e->getFriendlyMessage());
    return false;
} catch (\Mailchimp_Error $e) {
    $this->_helper->log($e->getFriendlyMessage());
    $this->_helper->markApiKeyFailed($storeId);
    return false;
}
```

So the transport case is excluded exactly as you wrote it.

**And the 404 case you raised is handled too, which your proposed fix does not reach.** A 404 for a wrong list id arrives as a plain `Mailchimp_Error` from `Mailchimp.php:312`, not as an `HttpError`, so ordering the catches leaves it recording a verdict. `_loadGroups()` therefore no longer records at all — it reads the verdict and stays silent, because the call it guards carries an audience id and its failure is ambiguous by construction. `_ping()` is the only request in the run that asks about the credential and nothing else.

That costs nothing in practice: both jobs share one cron process and the ecommerce job runs first — 10 of 10 consecutive runs on a live box, 1–2 seconds apart.

Since then the branch has been **rebased** onto `develop-2.4` (the two merges today conflicted in `Helper/Data.php`), so the SHA your review points at is no longer on the branch — the equivalent commit is `b5e41770`. That may be why it still reads as blocking.

I have also rewritten the PR description. It still described the original design, including the "deliberately broader than a 401" line I retracted in the comment above, which would have had you reading a description that contradicts the diff.

Tests now: twelve across three files, including one that fails with `Failed asserting that true is false` against the code before this change. Suite 157 tests, 280 assertions, green.

Nothing to do on my side that I can see — but if there is something in the current state you still want changed, say so and I will take it. Re-review when you get a chance.

dipola (approved):
Approving — and my change request should not have been there. You pushed `b5e4177` at 14:08; I submitted
at 15:08 from a diff I had fetched before it and never re-read. The fix was already in when I asked for it.
Sorry for the hour.

**And you went further than I asked, in the right direction.** I proposed ordering the catch in both
places. You ordered it in `Cron/Ecommerce.php` and then removed the verdict from `Cron/Webhook.php`
altogether, on the rule that only the call which asks about the credential *and nothing else* may condemn
it. That is a better rule than the one I gave you.

It is better because it makes the ambiguity structural rather than something a reader has to remember:
`lists->interestCategory->getAll($listId)` carries an audience id, so its failure is genuinely two claims
at once — bad key, or good key and wrong audience — and no amount of catch-ordering separates those.
`root->info()` carries nothing but the credential, so it is the only call in the run whose failure means
one thing. Reading `isApiKeyFailed` in the webhook loop while never writing it is exactly the right
asymmetry, and the comment at the catch says so plainly enough that the next person will not re-add it.

The tests are the part I would have missed:

- `testATransportFailureIsNotRecorded` — the case itself.
- `testAFailureHereNeverCondemnsTheCredential`, with `expects($this->never())->method('markApiKeyFailed')`
  as a class-level guard rather than an assertion about one call. That is what stops the verdict creeping
  back into the webhook loop in a year, which is the actual risk.
- `testAKnownFailedCredentialShortCircuits` asserting `1` account call across three views, not merely that
  the flag was set. It pins the saving rather than the mechanism.

The note in `ApiKeyVerdictTest` about modelling the double on a hash of the credential rather than on the
store — "or the test proves something the code does not do" — is the kind of thing that makes a test suite
worth having.

One expectation unchanged from my earlier comment, since it costs no code: this helps installs with many
store views on one key. It does not help a single-view install whose cron retries a dead key every run —
one rejected call per run either way, and the run frequency is the driver. Both shapes are in the field.

LGTM. Ship it with the others.

PR #2352, opened 2026-09-09
Name the store when loading Mailchimp stores

gonzaloebiz:
## The problem

`Helper\Data` has three places that hand an API key to the library. Two of them name the store; one does not.

| | `setApiKey` | `setHelper` | `setStoreURL` |
|---|---|---|---|
| `getApi()` | ✅ | ✅ | ✅ |
| `getApiByApiKey()` | ✅ | ✅ | ✅ |
| `loadStores()` | ✅ | ✅ | **missing** |

The asymmetry is visible as a bare count in the file: `setHelper` three times, `setStoreURL` twice.

`setApiKey()` is what opens the library's telemetry bucket, and `setStoreURL()` writes into an already-open bucket rather than keeping a copy on the instance. So a bucket opened without that call can never acquire a URL later — everything reported from that run is unattributable to a site, including the account id and subscriber count that this path is the only one to collect.

`loadStores()` has a single caller, `Controller/Adminhtml/Stores/Index.php:21`.

## The change

One call, after `setApiKey`:

```php
$this->_api->setStoreURL($this->_storeManager->getStore()->getBaseUrl());
```

Copied from the sibling in `getApiByApiKey()`, which is also admin-invoked and has always read the base URL this way — so this is existing shipped behaviour applied to a third site, not a new pattern.

## Verification

**Placement matters, and it was checked rather than assumed.** Driving the real library through each sequence and reading the bucket back:

| sequence | `store_url` in the bucket |
|---|---|
| `loadStores()` as it stands | `NULL` |
| with this call added, after `setApiKey` | `'https://…/'` |
| the same call placed *before* `setApiKey` | `NULL` |

The third row is why the line goes where it does: before `setApiKey` there is no bucket to write into, and the call is silently lost.

**Admin context returns the storefront URL, not the admin URL.** Worth confirming since the only caller is an admin controller — with the area set to `adminhtml`, `getStore()` returns the admin store (id 0) and `getBaseUrl()` still yields the storefront base URL.

Module unit suite: 138 tests, 249 assertions, green.

## Not done here

The four-line prelude now appears three times and has drifted once, so collapsing it into a shared private method is tempting. I left it alone deliberately: the three sites differ in four ways — one decrypts the key, one trims it, one takes a store parameter, one applies a timeout — so a shared helper would need parameters for each, and it would rewrite two working call sites to guard against a hypothetical fourth. Worth doing as its own change, with its own review, rather than riding along with a one-line fix.

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -1003,6 +1003,13 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
             $this->_api->setApiKey(trim($apiKey));
             $this->_api->setUserAgent('Mailchimp4Magento' . (string)$this->getModuleVersion());
             $this->_api->setHelper($this);
+            // Must stay after setApiKey. That call is what opens the library's
+            // telemetry bucket, and setStoreURL() writes into an already-open
+            // bucket rather than keeping a copy on the instance -- so moved
+            // above it this line is silently lost and the bucket reports with
+            // no site attached. Verified: before setApiKey the URL comes
+            // through NULL, after it comes through set.
+            $this->_api->setStoreURL($this->_storeManager->getStore()->getBaseUrl());
 
 
             try {


dipola:
Approving. This is the missing third call and it is in the right place.

The placement table is the most useful thing here — the third row, where the same call placed *before*
`setApiKey` leaves `store_url` NULL, is what turns "after `setApiKey`" from a preference into a
constraint. Worth keeping in the commit message so the next person moving lines around in that prelude
sees it.

The admin-context check answers the question a reviewer would otherwise have to ask: `getStore()` in
`adminhtml` returns store 0, and it is not obvious without checking that `getBaseUrl()` still yields the
storefront URL there rather than the admin one. That it matches the sibling in `getApiByApiKey()`, which
is also admin-invoked and has always read it this way, settles it.

Agreed on not bundling the `prepareApi()` refactor. The three sites differ more than the `setHelper`
count suggests, and rewriting two working call sites inside a one-line fix is the kind of thing that
makes a reviewer look at the wrong half of the diff. Happy to see it as its own PR if it is wanted.

gonzaloebiz:
Taken, and put in both places rather than only the one you asked for.

**The commit message** now carries the sequence table, so it survives in `git log` for whoever is reordering that prelude:

```
loadStores() as it stood            store_url = NULL
with the call, after setApiKey      store_url = the URL
the same call, before setApiKey     store_url = NULL
```

**And the code comment**, because that is what a person moving lines actually has in front of them. It used to explain why the call is needed; it now explains why it cannot move:

```php
// Must stay after setApiKey. That call is what opens the library's
// telemetry bucket, and setStoreURL() writes into an already-open
// bucket rather than keeping a copy on the instance -- so moved
// above it this line is silently lost and the bucket reports with
// no site attached.
```

Your framing is what made me change it: a commit message is where the reasoning is recorded, but the person about to break this is reading the file, not the history. The old comment would not have stopped them.

Amended and force-pushed — one commit, my branch, nobody else has it. Suite still green.

dipola (approved):
Approved. Detail in my comment above.

The one call, correctly placed after `setApiKey`. Your placement table is what turns "after setApiKey"
from a preference into a constraint — the row where the same call placed *before* it leaves `store_url`
NULL is worth keeping in the commit message for whoever next reorders that prelude.

Both pre-merge checks answered: `getStore()` in `adminhtml` returns store 0 but `getBaseUrl()` still
yields the storefront URL, and it matches the sibling in `getApiByApiKey()` which is also admin-invoked.

Agreed on not bundling the `prepareApi()` refactor here.

PR #2353, opened 2026-09-09
Report the Magento version in the user agent

gonzaloebiz:
## What it changes

The user agent sent with every Mailchimp API call has always been:

```
Mailchimp4Magento103.4.82
```

which reads as though the number were the Magento version. It is the **module's** `setup_version` — `getModuleVersion()` returns `$modules['Ebizmarts_MailChimp']['setup_version']`. The platform version was not carried anywhere.

It now reads:

```
Mailchimp4Magento103.4.82 Magento/2.4.8
```

39 characters against a 128 cap.

The slash is deliberate rather than cosmetic: it makes the two versions impossible to confuse for a reader, and it lets anything parsing the agent require `Magento/x.y.z` instead of guessing which number is which.

## Absent beats wrong

If `ProductMetadataInterface` cannot be asked, or answers with an empty string, the agent stays exactly as it was. Nothing invented, nothing empty appended — the change can only add a version that is really there.

## Three call sites became one method

`setUserAgent(...)` appeared three times in `Helper\Data`, byte-identical. Collapsing them was safe precisely because they were identical — unlike the surrounding four-line API prelude, which differs between sites and is deliberately left alone. A fourth caller would otherwise have had nothing to copy from but one of the three.

## Cost

`ProductMetadataInterface` was not injected in `Helper\Data`, so this adds a constructor argument — the 25th. Checked before doing it: nothing constructs the helper with `new` anywhere in the module, and all fourteen tests that touch it use `createMock`, so a new argument reaches none of them. Verified afterwards that the object manager still builds the helper.

## Verification

Six new tests, each watched failing against the unmodified helper first — all six error with `Call to undefined method` before the change and pass after.

They cover: both versions present; the slash prefix; the length staying under the cap; the fallback when the platform version is empty; the fallback when the lookup raises; and the module version still being findable inside the agent, since a cross-check that searches for it must not be broken by a second version appearing.

On a real install the method produces `Mailchimp4Magento103.4.82 Magento/2.4.8`.

Full module unit suite: 144 tests, 255 assertions, green.

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -15,6 +15,7 @@ use Magento\Framework\App\ResourceConnection;
 use Magento\Framework\Exception\ValidatorException;
 use Magento\Store\Model\Store;
 use Ebizmarts\MailChimp\Model\MailchimpNotificationFactory as MailchimpNotificationFactory;
+use Magento\Framework\App\ProductMetadataInterface;
 
 class Data extends \Magento\Framework\App\Helper\AbstractHelper
 {
@@ -135,6 +136,11 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
     const MIN_LIB_VERSION = '3.0.45';
 
     protected $counters = [];
+
+    /**
+     * @var ProductMetadataInterface
+     */
+    private $productMetadata;
     /**
      * @var \Magento\Store\Model\StoreManagerInterface
      */
@@ -291,7 +297,8 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
         \Magento\Framework\Stdlib\DateTime\DateTime $date,
         \Magento\Directory\Model\CountryFactory $countryFactory,
         \Magento\Framework\Locale\Resolver $resolver,
-        MailchimpNotificationFactory $mailchimpNotificationFactory
+        MailchimpNotificationFactory $mailchimpNotificationFactory,
+        ProductMetadataInterface $productMetadata
     ) {
 
         $this->_storeManager  = $storeManager;
@@ -319,6 +326,7 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
         $this->countryFactory           = $countryFactory;
         $this->resolver                 = $resolver;
         $this->mailchimpNotificationFactory = $mailchimpNotificationFactory;
+        $this->productMetadata          = $productMetadata;
         parent::__construct($context);
     }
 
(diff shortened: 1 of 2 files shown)


dipola:
Approving.

I checked the parsing side against the real agent string this produces, `Mailchimp4Magento103.4.82
Magento/2.4.8`. Both halves come back cleanly: an anchored match on the `Mailchimp4Magento` prefix stops
at the space and yields the module version, and a match requiring `Magento/` yields the platform version
without ever matching the `Magento` inside `Mailchimp4Magento`. The slash is doing exactly the work the
docblock says it is. Also checked `Mailchimp4Magento1.0.24 Magento/2.4.6-p15`, since a `-pN` suffix is the
common real-world shape — it parses too.

Absent-beats-wrong is the right default and it is implemented properly: both the `Exception` and
`Throwable` arms and the empty-string check all return the agent unchanged, so this can never emit
`Magento/` with nothing after it. That matters more than it looks — a trailing-slash-and-nothing value is
the shape that makes a parser at the other end produce an empty version rather than no version.

Two notes, neither blocking:

- On a PHP 7.4+ codebase `catch (\Throwable $t)` alone covers both arms, since `Exception` implements
  `Throwable`. Keeping both is harmless and matches the surrounding style, so this is only an observation.
- Adding the 25th constructor argument to the helper is worth the line, and your check that nothing
  constructs it with `new` and that every test uses `createMock` is the right thing to have done before
  rather than after. The "it was not one line" note is a good thing to have on the record.

gonzaloebiz:
Taking the `Throwable` note, not waving it through — I checked it and my reason for keeping both arms was wrong.

I had it in my head that the surrounding file used the pair. It does not: `Helper/Data.php` on `develop-2.4` has **five** `catch (Exception)` and **zero** `catch (\Throwable)`. The twelve `Throwable` catches in this module are all in `Model/Edge/`, not here. So the pair was not matching local style, it was introducing a new one — the opposite of the argument I would have given for it.

And the floor makes the second arm unreachable rather than merely redundant. `composer.json` declares no `php` constraint directly, but it requires `magento/framework: 103.0.*`, which is Magento 2.4 and therefore PHP 7.4 or newer. `Throwable` has existed since PHP 7.0, so nothing can reach the `Exception` arm that the `Throwable` one would not have caught first.

Collapsed to one, with the reason recorded where the next person will ask it:

```php
} catch (\Throwable $t) {
    // One arm covers both: Exception implements Throwable, and this
    // module's floor is Magento 2.4, so PHP 7 or newer is guaranteed.
    return $agent;
}
```

Amended and force-pushed — one commit, my branch. Suite green at 144 tests, 255 assertions, including the fallback test that exercises this exact path.

Thanks also for checking `Mailchimp4Magento1.0.24 Magento/2.4.6-p15`. The `-pN` shape is the one I would have missed, and it is the common real-world value rather than the clean `2.4.8` I tested against.

dipola (approved):
Approved. Detail in my comment above.

I checked the parsing side against the real agent this produces. Both halves come back cleanly: an
anchored match on the `Mailchimp4Magento` prefix stops at the space and yields the module version, and a
match requiring `Magento/` yields the platform version without ever matching the `Magento` inside
`Mailchimp4Magento`. Also checked a `-pN` platform version, since that is the common real-world shape.

Absent-beats-wrong is implemented properly in all three arms, so this can never emit `Magento/` with
nothing after it — which is the shape that makes a parser at the other end produce an empty version
rather than no version.

The "it was not one line" note is the most useful thing in the PR and I am glad it is on the record.

PR #2354, opened 2026-09-09
Stop resolving the connected-site URL on every page render

gonzaloebiz:
`Helper\Data::getJsUrl()` is called from `Block\Mailchimpjs`, which `view/frontend/layout/default.xml:17` places in `head.additional` on **every page**. So it resolves a URL by calling the Mailchimp API during page render — and two defects made it do that on every uncached render, indefinitely.

Full-page cache does not contain it: the block is not marked `cacheable="false"`, so an FPC hit costs nothing, but the call fires on every FPC miss — and on the pages this extension marks uncacheable itself (`checkout_cart_index`, `checkout_index_index`, the Hyvä checkout layouts, `mailchimp_cart_loadquote`). Every visitor to cart or checkout pays it regardless of FPC.

## 1. The admin placeholder went straight to the API

`Model/Config/Source/MonkeyStore.php:56` defines the dropdown's placeholder as `['value' => -1, 'label' => 'Select one Mailchimp Store']`. A merchant who never picks a store leaves `mailchimp/general/monkeystore` at `-1`, and that value was passed through unguarded:

```php
$storeData = $api->ecommerce->stores->get($mailChimpStoreId);   // "-1"
```

`GET /ecommerce/stores/-1` is a guaranteed 404. It was logged, nothing was saved, nothing was remembered, and the next render asked again.

The extension already knows `-1` means "nothing selected" — `Cron/Ecommerce.php:140` and `Model/Config/Source/Details.php:98` both guard it. This path did not.

## 2. The success path never flushed the config cache

The resolved URL was written with the raw resource model:

```php
$this->_config->saveConfig(self::XML_MAILCHIMP_JS_URL, $url, SCOPE_STORES, $storeId);
```

The helper's own `saveConfigValue()` (`:548`) does the same write **and then** `cleanType('config')`. Without that flush the config cache kept serving the old empty value, so `!$url` stayed true and the next render called the API again — a lookup that had succeeded and saved, and still never stopped asking.

This is the one with no symptom. It produces no errors at all, only latency, so it cannot be found by looking for failures.

## The change

Both fixes are inside `getJsUrl()`:

- return early when the store id is the `-1` placeholder, empty or null
- save through `saveConfigValue()` so the cache is flushed and the loop actually terminates

## Verification

Six tests, all watched failing against the unmodified code first:

| | before | after |
|---|---|---|
| `-1` as int | API called | **not called** |
| `-1` as string | API called | **not called** |
| empty / null store id | API called | **not called** |
| successful lookup | raw `saveConfig`, cache stale | saved via `saveConfigValue` |

The three placeholder tests fail with `Failed asserting that 1 is identical to 0` before the change — the 1 being the API call that should not have happened.

Also pinned: an already-resolved URL asks nothing, and a real store id that fails is logged and returns empty.

Full module unit suite: 144 tests, 263 assertions, green.

## Deliberately not fixed here

- **Failures are still not remembered.** A real store id that 404s is asked once per render. That needs somewhere to remember, with a TTL, and is a design decision rather than a guard.
- **Resolution still happens at render time.** A storefront render should not make an outbound API call at all; the value is already written at admin time by the connected-site provisioner. Moving the block to read-only is the right end state and a larger change than this.
- **`Model/Config/Backend/MonkeyStore::beforeSave` still persists `-1`.** Rejecting it there would stop the placeholder being stored in the first place, from the admin side.

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -1104,17 +1104,34 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
         $url = $this->getConfigValue(self::XML_MAILCHIMP_JS_URL, $storeId);
         if ($this->getConfigValue(self::XML_PATH_ACTIVE, $storeId) && !$url) {
             $mailChimpStoreId = $this->getConfigValue(self::XML_MAILCHIMP_STORE, $storeId);
+
+            // Two ways to hold a value that is not a store, and both reach
+            // here. -1 is the dropdown's placeholder, left behind by a
+            // merchant who never chose one; 0 is the '---No Data---' option
+            // MonkeyStore offers when there is no API key at that scope or the
+            // listing failed. This runs during page render, so without this
+            // guard every uncached visit pays to be told so again.
+            //
+            // 0 is the quieter of the two and worth the falsy check rather
+            // than another == comparison: the library branches on if($id), so
+            // a 0 does not even ask for that store -- it lists them, gets a
+            // 200 with no connected_site, saves nothing, and repeats with no
+            // error to notice it by.
+            if (!$mailChimpStoreId || $mailChimpStoreId == -1) {
+                return $url;
+            }
+
             try {
                 $api = $this->getApi($storeId);
                 $storeData = $api->ecommerce->stores->get($mailChimpStoreId);
                 if (isset($storeData['connected_site']['site_script']['url'])) {
                     $url = $storeData['connected_site']['site_script']['url'];
-                    $this->_config->saveConfig(
-                        self::XML_MAILCHIMP_JS_URL,
-                        $url,
-                        \Magento\Store\Model\ScopeInterface::SCOPE_STORES,
-                        $storeId
-                    );
+                    // Through the helper, which flushes the config cache. The
+                    // raw saveConfig used to leave the cache serving the old
+                    // empty value, so the very next render read !$url as true
+                    // and called again -- a successful lookup that never
+                    // stopped asking, and left no error behind to notice.
+                    $this->saveConfigValue(self::XML_MAILCHIMP_JS_URL, $url, $storeId);
                 }
             } catch (\Mailchimp_Error | \Mailchimp_HttpError $e) {
                 $this->log($e->getFriendlyMessage());
--- a/Test/Unit/Helper/JsUrlTest.php
+++ b/Test/Unit/Helper/JsUrlTest.php
@@ -1,0 +1,166 @@
+<?php
+/**
+ * Ebizmarts_MailChimp Magento Component
+ *
+ * @category    Ebizmarts
+ * @package     Ebizmarts_MailChimp
+ * @author      Ebizmarts Team <info@ebizmarts.com>
+ * @copyright   Ebizmarts (http://ebizmarts.com)
+ * @license     http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+namespace Ebizmarts\MailChimp\Test\Unit\Helper;
+
+use Ebizmarts\MailChimp\Helper\Data as MailChimpHelper;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * getJsUrl() runs during page render, so every call it makes is paid by a
+ * visitor. These pin the two ways it used to keep calling forever.
+ */
+class JsUrlTest extends TestCase
+{
+    /** @var int */
+    private $apiCalls = 0;
+
+    /** @var array */
+    private $saved = [];
+
+    /**
+     * @param  mixed $storeId   value of mailchimp/general/monkeystore
+     * @param  mixed $savedUrl  value already in config, or ''
+     * @param  mixed $answer    API answer, or a throwable
+     * @return MailChimpHelper
+     */
+    private function helper($storeId, $savedUrl, $answer = null)
+    {
+        $this->apiCalls = 0;
+        $this->saved    = [];
+
+        $api = new \stdClass();
+        $api->ecommerce = new \stdClass();
+        $api->ecommerce->stores = new class($answer, $this) {
+            private $answer;
+            private $test;
+            public function __construct($answer, $test) { $this->answer = $answer; $this->test = $test; }
+            public function get($id)
+            {
+                $this->test->countApiCall();
+                if ($this->answer instanceof \Throwable) { throw $this->answer; }
+                return $this->answer;
+            }
+        };
+
+        $helper = $this->getMockBuilder(MailChimpHelper::class)
+            ->disableOriginalConstructor()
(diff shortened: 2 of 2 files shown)


dipola (changes requested):
Both fixes are right, and Bug 2 being in here is the important half — a `-1` guard alone would have left
the succeeding population calling forever with no error signature, looking fixed while still costing
250 ms a render.

Requesting changes for one gap, because it creates a third population with that same invisible signature.

## The guard misses `0`, and `0` is reachable

`:1098` — `if ($mailChimpStoreId == -1 || $mailChimpStoreId === '' || $mailChimpStoreId === null)`.

`"0"` passes all three: `"0" == -1` is false, and it is neither `''` nor `null`. And `0` is a value the
dropdown really emits — `Model/Config/Source/MonkeyStore.php:71-73`:

```php
} else {
    $rc[] = ['value' => 0, 'label' => __('---No Data---')];
}
```

That is the **only** option whenever there is no API key at that scope (`:39`) or the store listing threw
(`:47`) — i.e. a fresh install's first save (Active=Yes, paste key, Save: the key was not there when the
page rendered), and any admin page load while Mailchimp is unreachable. `beforeSave` persists it.

**What happens then is worse than a 404.** `mailchimp-lib`'s `EcommerceStores::get()` branches on
`if($id)` (`EcommerceStores.php:65`), and `"0"` is falsy in PHP — so it does not request
`ecommerce/stores/0`, it requests the **list** endpoint `GET ecommerce/stores` (`:69`). That returns
**200**, carries no `connected_site` key, so `isset()` at `:1105` is false, nothing is saved, and the next
render repeats it. **Zero errors, forever** — the same shape as Bug 2, which is the one that took longest
to find precisely because nothing surfaces it.

One token fixes it:

```php
if (!$mailChimpStoreId || $mailChimpStoreId == -1) {
```

`!$mailChimpStoreId` subsumes `''`, `null`, `'0'` and `0`; the `== -1` keeps the placeholder. Worth a
`'0'` case in the test table beside the existing three.

**Calibration, so this is weighted properly:** I checked our telemetry and there is no install sitting at
`0` today — every affected one is on `-1`. So this is a latent path, not a live fire, and it should not
hold the release hostage if you would rather ship and follow up. I am asking for it now only because it is
one token and its failure mode is the invisible one.

Small thing for the PR body rather than the code: `''` and `null` were not "a guaranteed 404" before this
either — same falsy branch, so they were a successful list call per render. The PR fixes that too; the
description just misdescribes what it was.

## A real store id that 404s still loops — and that is fine for this PR

Traced it: id passes the guard, `stores->get()` throws `Mailchimp_Error`, `:1114` catches, `log()` is a
no-op unless logging is on, `$url` stays null, nothing is written, next render repeats. Your own
`testARealStoreThatFailsIsLoggedAndReturnsEmpty` pins that as intended, which I read as deliberate scope
rather than an oversight.

Same for a store that exists but has no connected site — which the dropdown explicitly offers as
*"(Warning: not connected)"* (`Source/MonkeyStore.php:65`) — and for a Mailchimp timeout, where every
uncached render blocks for the curl timeout before serving the page. That last one is the widest blast
radius of the three and it is not merchant error at all.

I have no observed installs in any of those states right now, so nothing is burning. But it is the
argument for taking your item (3), the negative cache, **first** rather than last: it is the only one that
covers them, and it is the difference between "misconfigured merchants loop" and "any unresolved store
loops, including during a Mailchimp outage".

On the rest of your ordering — (5) protects no additional install from the loop now that the read side is
guarded; its value is stopping the null list id it writes when its own lookup fails, which is worth doing
on its own merits. And the template change protects nobody: an empty `src` fetches nothing either way. It
is a one-line rider for whichever PR next touches that file, not a priority.

`saveConfigValue()` is the right call and writes to the same scope and id as the raw one it replaces —
I read the default branch rather than trusting the mock. No regression risk found: the early return hands
back the same `$url` the old catch path fell through to, so the block and template render identical output
for every input.

gonzaloebiz:
Confirmed and taken. Your guard, not mine — it is better in both directions.

## The `0` path is real, and my guard let it through

Verified rather than accepted:

- `Model/Config/Source/MonkeyStore.php:72` emits `['value' => 0, 'label' => '---No Data---']` in the `else` branch, so it is reachable whenever there is no API key at that scope or the listing threw.
- My condition — `== -1 || === '' || === null` — **passes `0` and `'0'` straight through**. Checked each value rather than reasoned about it.

And your point about it being worse than the 404 holds: `EcommerceStores::get()` branches on `if ($id)`, and `0` is falsy, so it does not ask for that store at all. It **lists** them, gets a 200 with no `connected_site`, saves nothing, and repeats every render with no error to notice it by — Bug 2's signature reached by a different road.

`if (!$mailChimpStoreId || $mailChimpStoreId == -1)` covers `0`, `'0'`, `''` and `null` in one falsy check, keeps `-1`, and still lets a real id through. Confirmed `'abc123'` passes both before and after.

**The test was watched failing against my own previous guard**, not only against the original code:

```
Failed asserting that 1 is identical to 0
```

— the 1 being the API call my guard did not prevent. Suite: 151 tests, 273 assertions, green.

## On the calibration

Noted, and it matches what I would have wanted to know: latent, nothing sitting at `0` today, and not a reason to hold anything. It is in because it costs one condition and the same PR is already touching that line.

## On your correction

Thanks for making it. It changes what this PR is: the "real store id that 404s" bucket was `mc_store_id` NULL and there is nobody in it, so the remaining unfixed path — a real id that fails, asked once per render — is a code path with no occupants rather than a live cost. That moves "remember failures" further down the list than I had it, and I would not now argue for doing it before moving the resolution out of render time.

Worth having on the record here so the next reader does not inherit the urgency I attached to it.

dipola (approved):
Approving — clearing my own change request, because I had the release model wrong.

I blocked this on the `0` guard while assuming the release itself was somebody else's call and therefore
far off. You cut the releases here, so the trade I actually made was: hold a fix for four live merchants,
whose storefront tracking has never worked and who pay ~250 ms on every uncached render, behind a latent
path that **no install is currently sitting in**. That is the wrong way round, and it was my error rather
than a judgement call.

The `0` finding stands and is worth doing — it is one token
(`if (!$mailChimpStoreId || $mailChimpStoreId == -1)`), and its failure mode is the invisible one: `"0"` is
falsy in `EcommerceStores::get()`, so it calls the LIST endpoint, gets a 200 with no `connected_site`,
saves nothing and repeats every render with no error at all. But it belongs in a follow-up, not in front of
this.

Everything else I checked stands: `saveConfigValue()` writes to the same scope and id as the raw call it
replaces, and the early return hands back the same `$url` the old catch path fell through to, so the block
and template render identical output for every input. No regression found.

Ship it whenever suits. The sooner it is in a release the sooner those eleven store views stop calling
Mailchimp on every page view — nothing on our side changes that for them, it only stops us paying for the
noise and lets the console name the fault.

PR #2355, opened 2026-09-09
Fetch interest groups once per store view, and only when they can be used

gonzaloebiz:
Three ways the same interest groups were fetched over and over. None of them leaves an error behind, which is why none had been noticed.

## 1. An empty result was read as "nothing was supplied"

`getSubscriberInterest()` takes the groups as an optional third argument and used `!$interest` to decide whether the caller had passed them:

```php
public function getSubscriberInterest($subscriberId, $storeId, $interest = null)
{
    if (!$interest) {
        $interest = $this->getInterest($storeId);
    }
```

`[]` is falsy. And `[]` is exactly what `getInterest()` returns for a store with no groups configured — **on a completely successful call**. So the supplied value was treated as a cache miss and the fetch re-issued, once per subscriber, forever, with no error and no log line.

`Model/Api/Subscriber.php` is where that multiplies: it primes the groups, then passes them back in for every subscriber row.

Now `$interest === null`, matching the parameter's own default. The three callers that genuinely supply nothing pass `null` and are unaffected.

## 2. The call was made when its answer could not be used

With no groups selected, `getInterest()` still asked Mailchimp — then discarded the answer, because the first loop matches category ids against an empty list and the second iterates that same empty list. The result is `[]` however the call goes.

That is the quiet majority rather than an edge case: an audience configured and no groups chosen is the default state of any install that never used them.

The audience id had the same placeholder problem the store id did, too — `-1` from the dropdown, or an empty value building `lists//interest-categories`. `Cron/Webhook.php:351` already guards this exact call that way; this path did not.

## 3. The groups were fetched before there was anything to sync

`sendSubscribers()` fetched them as its first statement, before the subscriber collection was even built, and `_processStore()` calls it unconditionally. So every store view of every run paid for the groups whether or not a single subscriber was waiting.

They are now loaded on first use, with a flag of their own — `[]` has to be distinguishable from "not fetched yet", which is the same distinction defect 1 got wrong.

## The trap that shaped the fix

`Model\Api\Subscriber` has **no `shared="false"` in `etc/di.xml`**, so one instance serves every store view in a cron run — confirmed against the object manager, including that `Cron\Ecommerce` holds that same instance.

A loaded-flag living on the instance would therefore have served store view 1's groups to store view 2: wrong groups synced, no error, no log line. Exactly the class of defect this PR is about.

Both the groups and the flag are reset at the top of `sendSubscribers()`, so the memory is scoped to the view being synced.

## Verification

Nine tests across two files, all watched failing against the unmodified code first:

| | before | after |
|---|---|---|
| no groups configured | 1 call | **0** |
| audience id `''`, `null`, `0`, `'0'`, `-1`, `'-1'` | 2 calls | **0** |
| `[]` supplied by the caller | 2 calls | **0** |
| five subscribers in one view | 5 fetches | **1** |
| a view with no subscribers | 1 fetch | **0** |
| two views in one run | view 2 inherits view 1's groups | each fetches its own |

Full module unit suite: 173 tests, 320 assertions, green.

## Not changed

`Model/Config/Source/MonkeyList` still cannot enumerate the account's audiences, so the dropdown offers only what is already saved. That is a behaviour change and belongs in its own PR — and it is why `required-entry` must not be added to that field, which would make the section unsavable.

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -1296,9 +1296,27 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
         } else {
             $interest = [];
         }
+
+        // Nothing below can use the answer when no groups are selected: the
+        // first loop matches against an empty list and the second iterates
+        // one, so $rc comes back empty however the call goes. This is the
+        // quiet majority -- an audience configured and no groups chosen is
+        // the default state of an install that never used them.
+        if (!$interest) {
+            return $rc;
+        }
+
+        // And the audience id has the same placeholder problem as the store
+        // id: -1 is the dropdown's, and an empty value builds
+        // `lists//interest-categories`. Cron/Webhook guards this exact call
+        // the same way.
+        $listId = $this->getConfigValue(self::XML_PATH_LIST, $storeId);
+        if (!$listId || $listId == -1) {
+            return $rc;
+        }
+
         try {
             $api = $this->getApi($storeId);
-            $listId = $this->getConfigValue(self::XML_PATH_LIST, $storeId);
             $allInterest = $api->lists->interestCategory->getAll($listId, null, null, 200);
             if (is_array($allInterest) &&
                 array_key_exists('categories', $allInterest) &&
@@ -1327,7 +1345,12 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
     }
     public function getSubscriberInterest($subscriberId, $storeId, $interest = null)
     {
-        if (!$interest) {
+        // Strictly null, because [] is an answer and not an absence. Read as
+        // falsy it meant "the caller supplied nothing", so a store with no
+        // groups configured -- where this legitimately returns [] -- re-asked
+        // Mailchimp once per subscriber, on a successful call, with nothing to
+        // show for it and no error to notice it by.
+        if ($interest === null) {
             $interest = $this->getInterest($storeId);
         }
         /**
--- a/Model/Api/Subscriber.php
+++ b/Model/Api/Subscriber.php
@@ -40,6 +40,15 @@ class Subscriber
     protected $_interest=null;
 
     /**
+     * Whether the interest groups for the store view being synced have been
+     * fetched. Separate from $_interest because [] is a real answer: without
+     * it, a store with no groups configured looks unfetched forever.
+     *
+     * @var bool
+     */
+    protected $_interestLoaded = false;
+
+    /**
      * @param \Ebizmarts\MailChimp\Helper\Data $helper
      * @param SyncHelper $syncHelper
      * @param \Magento\Newsletter\Model\ResourceModel\Subscriber\CollectionFactory $subscriberCollection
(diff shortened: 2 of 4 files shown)


dipola (approved):
Approving. I checked the two things that could have made this wrong, and both hold.

**The early return is provably equivalent, not just probably.** I read the whole of `getInterest()` on `develop-2.4` rather than the hunk. With `$interest === []` the first loop's `in_array($item['id'], $interest)` can never match and the second loop `foreach ($interest as $interestId)` never executes, so `$rc` is `[]` on every path through the method. Nothing else in there has a side effect — the only other statements are the `log()` on a malformed response and the `log()` in the catch. So skipping the call cannot change the return value, and the only observable difference is two log lines that described a call we are no longer making.

**The staleness trap you identified is genuinely closed.** This was my main concern, because a loaded-flag on a DI singleton is exactly the shape of the defect being fixed. Call graph on `develop-2.4`:

```
_getInterest()          <- only caller: _buildSubscriberData():133
_buildSubscriberData()  <- only caller: sendSubscribers():89
```

`update()` and `deleteSubscriber()` are the other two public entry points and neither touches `_interest` or `_getInterest`. So `sendSubscribers()` is the sole path in, the reset is its first statement, and there is no way to reach the lazy load with a flag left over from another store view. Keying the load on `$subscriber->getStoreId()` is also safe here specifically because the collection is filtered `store_id = $storeId`, so every row in the loop carries the id the reset was performed for.

`!$listId || $listId == -1` — the loose compare is the right call. It is what `Cron/Webhook.php:351` already does, and having the two copies of this guard read identically is worth more than tightening one of them in isolation. It behaves correctly for `'-1'` and for non-numeric ids on both PHP 7 and 8.

---

**One thing that is not blocking, but the summary table claims more than the tests prove.**

Two of the rows are verified by a test double of the production reset rather than by the reset itself. `beginStore()` sets `_interest = null` and `_interestLoaded = false` through reflection — which re-implements the two lines at the top of `sendSubscribers()` instead of calling them. Consequences:

- `testOneStoreViewDoesNotInheritAnothersGroups` is the test named after the trap, and it would still pass if someone deleted both reset lines from `sendSubscribers()`. The behaviour it guards is the one thing here that fails silently, so it is the one I would most want pinned to the real code.
- `testAStoreViewWithNoSubscribersFetchesNothing` calls only `beginStore()` and then asserts nothing was fetched — it is asserting that not calling `_getInterest()` does not call `getInterest()`. The actual claim, that `sendSubscribers()` no longer fetches before the collection is built, is not exercised.

I understand why: `sendSubscribers()` drags in the collection factory, `syncHelper` and `getTableName()`, and that is a lot of wiring for two assertions. So I am not asking you to hold the release for it. But the row that says *"two views in one run: view 2 inherits view 1's groups -> each fetches its own"* is currently guaranteed by the test helper rather than by the module, and that is worth knowing when the next person changes `sendSubscribers()`.

The other seven tests do exercise real code and are the ones carrying the weight — `testAnUnsetAudienceAsksNothing` iterating all six placeholder shapes is the right level of paranoia for a value that arrives from a dropdown.

Ship it.

dipola:
Merged — thanks. That was the last of today's batch, so from my side there is nothing else waiting on review.

**Please cut 103.4.82 now rather than letting it accumulate further.** `develop-2.4` is carrying **16 merged PRs** that have never been tagged, going back to 103.4.81 on 2026-08-04 — five weeks:

```
#2338 #2340 #2341 #2342 #2343 #2344 #2346 #2347
#2349 #2350 #2351 #2352 #2353 #2354 #2355
```
(plus #2348, which is contributor-guide only)

The usual shape: `version-82-2.4` off `develop-2.4`, touching `CHANGELOG.md`, `composer.json` and `etc/module.xml`, the way #2337 did for 81.

The reason I am pushing on the timing rather than just the content: **five of today's merges only do anything once a merchant can install them.** #2351, #2352, #2353, #2354 and #2355 all reduce Mailchimp API calls made from merchant servers — a rejected key asked once per run instead of once per store view, the connected-site URL no longer resolved on every page render, interest groups fetched once per store view instead of once per subscriber. Every one of those is a saving that stays theoretical while it sits on an untagged branch. The merchants currently paying that cost are on 103.4.81 and will keep paying it until a tag exists.

Nothing else is blocking. The remaining `getJsUrl` follow-ups are tracked separately, they are not in this release, and they should not hold it.

PR #2356, opened 2026-09-09
Pin the interest reset to sendSubscribers, not to the test helper

gonzaloebiz:
Follow-up to PR 2355, raised in review there and worth fixing rather than leaving.

## The test proved less than it claimed

`SubscriberInterestTest` began a store view like this:

```php
private function beginStore(Subscriber $api, $storeId)
{
    // reset both fields directly, by reflection
}
```

So `testOneStoreViewDoesNotInheritAnothersGroups` — the one named after the singleton trap — **passed whether or not the module reset anything**. Deleting both reset lines from `sendSubscribers()` left it green. It was pinning the test helper.

That matters more than usual here, because the row in 2355's table reading *"view 2 inherits view 1's groups → each fetches its own"* was, as written, guaranteed by the fixture rather than by the code.

## What it does now

A view begins the way the cron begins one: by calling `sendSubscribers()`, with a collection double that yields nothing. An empty view is the right case to drive it with — it must still reset, and it must still cost no fetch.

## Verified by breaking it

With the two reset lines removed from `Model/Api/Subscriber.php`:

```
1) SubscriberInterestTest::testOneStoreViewDoesNotInheritAnothersGroups
Failed asserting that two arrays are identical.
Tests: 4, Assertions: 4, Failures: 1.
```

Restored, green. That failure is the whole point of the change — before it, the same deletion was silent.

Full module unit suite: 173 tests, 320 assertions, green.

## Scope

Test only. No production file is touched.

### DIFF
--- a/Test/Unit/Model/Api/SubscriberInterestTest.php
+++ b/Test/Unit/Model/Api/SubscriberInterestTest.php
@@ -44,31 +44,58 @@ class SubscriberInterestTest extends TestCase
             }
         );
 
+        $helper->method('getTableName')->willReturn('mailchimp_sync_ecommerce');
+        $helper->method('getDateMicrotime')->willReturn('1757000000');
+
+        // Enough of a collection for sendSubscribers() to walk it and find
+        // nothing. An empty view is the case that matters here: it must still
+        // reset, and it must still cost no fetch.
+        $select = new class {
+            public function joinLeft() { return $this; }
+            public function where() { return $this; }
+            public function limit() { return $this; }
+        };
+        $collection = new class($select) implements \IteratorAggregate {
+            private $select;
+            public function __construct($select) { $this->select = $select; }
+            public function addFieldToFilter() { return $this; }
+            public function getSelect() { return $this->select; }
+            public function getIterator(): \Traversable { return new \ArrayIterator([]); }
+        };
+        $collectionFactory = $this->getMockBuilder(\stdClass::class)
+            ->addMethods(['create'])
+            ->getMock();
+        $collectionFactory->method('create')->willReturn($collection);
+
         $api = $this->getMockBuilder(Subscriber::class)
             ->disableOriginalConstructor()
             ->onlyMethods([])
             ->getMock();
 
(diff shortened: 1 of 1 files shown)


dipola (approved):
Approving. This is the right fix and it is verified the only way this particular class of defect can be verified — by deleting the production reset and watching the test go red. A test that has never been observed failing for the reason it exists is not yet evidence, and now this one has been.

Checked the fixture actually lets `sendSubscribers()` run to completion rather than dying early somewhere and passing for the wrong reason:

- the collection chains — `addFieldToFilter()` twice returning `$this`, then `getSelect()->joinLeft()`, `->where()`, `->limit()` — all resolve, and PHP lets those no-arg anonymous methods take the real call's arguments and ignore them
- `getTableName()` and `getDateMicrotime()` are stubbed, which are the two helper calls made before the loop
- the empty `ArrayIterator` means the row loop never runs, so `modifyCounter()`, `_updateSubscriber()` and `syncHelper` are never reached — which is why leaving `syncHelper` unset is safe rather than lucky

The second test gets fixed for free and that is worth naming, because it was the weaker of the two. `testAStoreViewWithNoSubscribersFetchesNothing` previously asserted that not calling `_getInterest()` did not call `getInterest()` — true of any code at all. Now it calls `sendSubscribers()` with a collection that yields nothing, so it actually pins the claim it is named after: that the fetch no longer happens before the collection is built.

> both times it was a *helper* in the test that did the work the production code was supposed to be doing

That is the generalisation, and it is a sharper one than "the assertion was vacuous". The failure is not carelessness in the assertion — it is that setup code and code under test were doing the same job, so the test could not tell them apart. The check that catches it is cheap and mechanical: break the production line the test is named after, and require the failure before believing the pass.

One small thing, not worth changing: the anonymous collection satisfies no interface, so if `sendSubscribers()` ever calls a method it does not have, this fails as a fatal rather than as a readable assertion. Acceptable at this size — worth remembering only if that method grows.

Correctly kept out of 103.4.82. Test-only, no reason to hold the tag for it.

PR #2357, opened 2026-09-09
version 82 for magento 2.4

gonzaloebiz:
Release preparation for 103.4.82, following the shape of #2337.

## Only the changelog changes

`composer.json` and `etc/module.xml` **already carry 103.4.82** — the bump arrived with #2350 rather than in a release PR — so unlike #2337 there is nothing to bump here. Verified both before writing this.

## Seventeen pull requests, not sixteen

`git log 103.4.81..develop-2.4` gives:

```
2338 2340 2341 2342 2343 2344 2345 2346 2347 2348
2349 2350 2351 2352 2353 2354 2355
```

That is five weeks of work — 103.4.81 was cut on 2026-08-04.

Worth flagging one: **#2345** — *"Write a broadcast notification to the inbox once, not once per store view"* — is a functional fix and was missing from the list this was requested against. It is in the entry.

#2356 is open and is test-only, so it is not in this release.

## Grouping

The entry keeps the file's existing headings, with one addition: **"Fewer calls to the Mailchimp API from merchant servers"** for #2351, #2354 and #2355. They are bug fixes in the ordinary sense, but grouping them says the thing a merchant reading the changelog would want to know — that this release reduces what the extension asks of Mailchimp from their server, on every page render and on every cron run.

## After this merges

The tag is what actually delivers any of it. Until `103.4.82` exists, everything above stays on an untagged branch and no merchant can install it.

Suite: 173 tests, 320 assertions, green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,38 @@
 # Changelog
 
+## [103.4.82](https://github.com/mailchimp/mc-magento2/tree/103.4.82)
+
+[Full Changelog](https://github.com/mailchimp/mc-magento2/compare/103.4.81...103.4.82)
+
+**Implemented enhancements:**
+
+- Report extension status to the ebizmarts service once an hour [\#2341](https://github.com/mailchimp/mc-magento2/pull/2341)
+- Let merchants opt out of sharing account contact details [\#2344](https://github.com/mailchimp/mc-magento2/pull/2344)
+- Tell the library which Mailchimp store this store view syncs into [\#2350](https://github.com/mailchimp/mc-magento2/pull/2350)
+- Report the Magento version in the user agent [\#2353](https://github.com/mailchimp/mc-magento2/pull/2353)
+
+**Fewer calls to the Mailchimp API from merchant servers:**
+
+- Stop resolving the connected-site URL on every page render [\#2354](https://github.com/mailchimp/mc-magento2/pull/2354)
+- Fetch interest groups once per store view, and only when they can be used [\#2355](https://github.com/mailchimp/mc-magento2/pull/2355)
+- Ask a rejected API key once per run, not once per store view [\#2351](https://github.com/mailchimp/mc-magento2/pull/2351)
+
+**Fixed bugs:**
+
+- Fix profile webhook recreating newsletter subscribers as Not Activated [\#2338](https://github.com/mailchimp/mc-magento2/pull/2338)
+- Send the chosen variant in ProductVariantDto.id for configurable products [\#2340](https://github.com/mailchimp/mc-magento2/pull/2340)
+- Write a broadcast notification to the inbox once, not once per store view [\#2345](https://github.com/mailchimp/mc-magento2/pull/2345)
+- Bound mailchimp_errors with a row ceiling, without overriding anyone [\#2347](https://github.com/mailchimp/mc-magento2/pull/2347)
+- Apply the Error fix to the rest of the edge classes [\#2349](https://github.com/mailchimp/mc-magento2/pull/2349)
+- Name the store when loading Mailchimp stores [\#2352](https://github.com/mailchimp/mc-magento2/pull/2352)
+
+**Merged pull requests:**
+
+- Keep tests out of release archives [\#2342](https://github.com/mailchimp/mc-magento2/pull/2342)
+- Update PixelStateWriter tests to the behaviour that shipped [\#2343](https://github.com/mailchimp/mc-magento2/pull/2343)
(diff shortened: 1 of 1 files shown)


dipola (approved):
Approving. Checked against the commit range rather than against the list I gave you, since that list was the thing you corrected.

**You are right that it is seventeen, and my sixteen was wrong.** I built my count from the merged-PR list filtered by base branch, which cannot see #2345 — it merged into `feat/edge-beacon-register-ping-notifications`, and that branch later landed on `develop-2.4`, so the commits are in the range while the PR's own base is not. `git log 103.4.81..develop-2.4` is the correct instrument and mine was not. Worth remembering, because a functional fix hiding behind a feature-branch merge is exactly the entry a release note drops silently.

**The entry is complete and exact.** Extracted every PR number from the diff and compared it to the same range:

```
in changelog: 17    in range: 17
missing from changelog: (none)
in changelog but not in range: (none)
duplicated: (none)
```

**Nothing to bump, confirmed independently.** `composer.json` on `develop-2.4` reads `103.4.82` where the 103.4.81 tag reads `103.4.81`, and `etc/module.xml` carries `setup_version="103.4.82"`. So the divergence from #2337's shape is correct, not an omission.

**And a correction of my own.** I said in my last comment that five of today's merges reduce API calls from merchant servers. Three do — #2351, #2354 and #2355, which is exactly the set you grouped. #2352 names the store when loading and #2353 adds the Magento version to the user agent; neither removes a call. Your grouping is the accurate one and mine was loose.

The new heading earns its place for the same reason: those three are ordinary bug fixes by mechanism, but what a merchant wants from a changelog is what changes on their server, and "your store stops asking Mailchimp for the same thing on every page render" is that. Keeping it separate from **Fixed bugs** is right.

One cosmetic thing only, take it or leave it: the three entries under the new heading run 2354, 2355, 2351 while every other group ascends.

Agreed on the last line — the tag is the delivery. None of this reaches anyone until `103.4.82` exists.

gonzaloebiz:
Sorted — the three now run 2351, 2354, 2355, and I checked all four groups rather than just the one you pointed at:

```
Implemented enhancements                                [2341, 2344, 2350, 2353]  ascending
Fewer calls to the Mailchimp API from merchant servers  [2351, 2354, 2355]        ascending
Fixed bugs                                              [2338, 2340, 2345, 2347, 2349, 2352]  ascending
Merged pull requests                                    [2342, 2343, 2346, 2348]  ascending
```

I had ordered that group by size of effect, which is a rule the file does not use anywhere else. Consistency wins.

## On #2345

Your explanation is the useful part and I had not worked out the mechanism — I only knew the two lists disagreed. A PR whose own base is a feature branch is invisible to a base-filtered list while its commits are squarely in the range, so the release note drops it silently and nothing anywhere says so. That is worth more than the one entry it saved here.

## On your correction

Taking it: #2352 and #2353 do not remove a call. #2352 attaches a store URL to a beacon that already went out, #2353 lengthens a header. Neither changes what a merchant's server asks Mailchimp for. Three, not five.

Amended and force-pushed — one commit, my branch. Ready when you are, and the tag is the part that actually delivers it.

dipola:
Approved about fourteen hours ago and `develop-2.4` is still untagged — please merge this and push `103.4.82`.

One measured reason that arrived overnight, from a live install running 103.4.81. Its cron telemetry is byte-identical on every run: 25 API calls, 41 seconds, ~232 KB downloaded, and **zero bytes ever uploaded**. Almost all of it is `Cron\Webhook::_loadGroups()`, which runs as the first statement of `processWebhooks()` — before the webhook queue is read — and whose result is consumed only by `profile`-type rows. On the `*/5` schedule that is over six thousand discarded `lists` calls a day on one merchant. I will raise that separately; it is not a blocker for this PR.

It is a reason to tag today rather than later, though: #2351, #2354 and #2355 all reduce exactly this kind of traffic, and none of them reaches a single merchant until the tag exists. The merchants paying that cost right now are on 103.4.81 by definition.

#2356 is test-only and correctly excluded — nothing here should wait for it.

dipola:
Approved about fourteen hours ago and `develop-2.4` is still untagged — please merge this and push `103.4.82`.

One reason to tag today rather than later. `Cron\Webhook::_loadGroups()` runs as the first statement of `processWebhooks()` — before the webhook queue is read — and its result is consumed only by `profile`-type rows. On the `*/5` schedule an install with nothing of that type queued still pays the full interest-category fetch every five minutes, indefinitely. I will raise that separately; it is not a blocker for this PR.

But it is the reason the tag matters: #2351, #2354 and #2355 all reduce exactly this kind of traffic, and none of them reaches anyone until the tag exists.

#2356 is test-only and correctly excluded — nothing here should wait for it.

PR #2358, opened 2026-09-10
Remember a failed connected-site lookup instead of repeating it every render

gonzaloebiz:
Follow-up to the connected-site work in PR 2354, covering what that one deliberately left out.

## The problem

2354 returns early for `-1`, `''`, `null`, `0` and `'0'`, so the misconfigured loop is closed. What still runs on **every uncached page render** is a store id that is real but does not resolve: it passes the guard, `stores->get()` throws or answers without `connected_site.site_script.url`, the catch logs (a no-op unless logging is on), `$url` stays empty, nothing is written, and the next render repeats it.

Four ways to get there:

- the Mailchimp store was **deleted** on Mailchimp's side
- a **restored database** carries a store id belonging to another account
- the store exists but has **no connected site** — which the dropdown itself offers as *"(Warning: not connected)"*
- the **key was revoked**, so the call 401s

And the one that is not merchant error at all: while **Mailchimp is down or slow**, every uncached render of every store using the pixel blocks on curl until the library's timeout before the page is served — cart and checkout included, since the extension marks those uncacheable itself. That case is bounded by nothing we control, which is why the other three are worth fixing alongside it.

## The change

A failure is remembered for fifteen minutes, so an unresolved store is asked about once per window instead of once per visitor.

The marker is written in **two** places, not one:

- in the `catch` — which is what covers the outage, because the library raises `Mailchimp_HttpError` on a curl failure as well as on an HTTP one (`if(curl_error($ch)) throw new Mailchimp_HttpError`), so the timeout lands in the same handler as the 404 and the 401;
- in a new `else` for the **200 that carries no URL** — the quietest failure of the set, where nothing throws, nothing is logged and nothing is written, so it would otherwise keep asking forever with no error to notice it by.

`MonkeyStore::beforeSave()` and the "Fix Mailchimp JS" button both drop the markers, so an admin fix takes effect on the next render rather than after the window. The TTL therefore only governs repairs made on Mailchimp's side — a site reconnected, a key restored — which is what fifteen minutes is chosen for.

`clearJsUrlFailures()` cleans by tag rather than per store view on purpose: in `beforeSave` the scope id can be a website or `0`, while the marker is keyed by store view id, so being surgical there would simply be wrong. Both call sites are rare, deliberate admin actions, and the cost of dropping all the markers is at most one call per store view on the next render — which is what those store views paid on *every* render before this.

## Why the cache argument is in di.xml

The argument is optional so the constructor stays backwards compatible, and that optionality is exactly why it has to be named in `di.xml`: `ClassReader` records a parameter with a default value as **not required**, and `AbstractFactory::getResolvedArgument()` hands a not-required parameter its default instead of resolving its type. Without that line the argument arrives `null` on every instantiation and this change does nothing at all — silently, with the unit tests still green, because they inject the double themselves. There is a test pinning both halves.

## Verified

Unit suite: 183 tests, 357 assertions. The new tests were each checked against a mutant — removing the read guard, the marker in the `catch`, the marker in the 200-without-URL branch, the lifetime, or the tag all produce failures.

Also measured on a 2.4.8 install with `monkeystore` set to a 32-hex id that 404s, no saved URL, and genuinely uncached renders (unique query string past Varnish), counting calls by entries added to `var/log/MailChimp.log`:

| 4 uncached renders | API calls | render times |
|---|---|---|
| before | **4** | 1.87 / 1.14 / 1.24 / 1.31 s |
| after | **1** | 1.16 / 0.39 / 0.41 / 0.35 s |

At helper level across separate processes, which is what shows the marker surviving a request: first call 0.788s, second and third 0.000s; after `clearJsUrlFailures()`, 0.642s again and back to 0.000s.

Note for testing: the di.xml change needs a `cache:flush` before it takes effect, since the merged DI config lives in the `config` cache type even in developer mode. A version bump's `setup:upgrade` covers that for merchants.

### DIFF
--- a/Controller/Adminhtml/Ecommerce/FixMailchimpJS.php
+++ b/Controller/Adminhtml/Ecommerce/FixMailchimpJS.php
@@ -72,6 +72,11 @@ class FixMailchimpJS extends \Magento\Backend\App\Action
         $resultJson = $this->resultJsonFactory->create();
         try {
             $this->config->deleteConfig(\Ebizmarts\MailChimp\Helper\Data::XML_MAILCHIMP_JS_URL, $scope, $scopeId);
+            // This button exists to make the next render look the URL up
+            // again. Dropping the config value alone would not do that for a
+            // store view that had been failing: it would still be holding a
+            // marker telling the helper not to ask.
+            $this->helper->clearJsUrlFailures();
         } catch(ValidatorException $e) {
             $valid = 0;
             $message = $e->getMessage();
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -92,6 +92,28 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
     const XML_PIXEL_SCRIPT_URL        = 'mailchimp/pixel/script_url';
     const XML_PIXEL_SCRIPT_FRAGMENT   = 'mailchimp/pixel/script_fragment';
 
+    /**
+     * How long a failed connected-site lookup is remembered for.
+     *
+     * Fifteen minutes is the balance between the two ways this can be wrong.
+     * Too short and a Mailchimp outage is still paid for by visitors, since
+     * the whole point is that a store view attempts once per window instead of
+     * once per render. Too long and a merchant who fixes the cause outside
+     * Magento -- reconnecting the site, restoring a revoked key -- waits for
+     * the window to run out, because nothing inside Magento changes to say so.
+     *
+     * The admin's own fixes do not wait: choosing a different Mailchimp store
+     * and the "Fix Mailchimp JS" button both drop the markers outright, so the
+     * TTL only ever governs repairs made on Mailchimp's side.
+     */
+    const JS_URL_FAILURE_TTL         = 900;
+
+    /**
+     * Tagged so those two admin paths can drop every marker in one call,
+     * without having to work out which store views hold one.
+     */
+    const JS_URL_FAILURE_CACHE_TAG   = 'MAILCHIMP_JS_URL_FAILURE';
+
     const ORDER_STATE_OK             = 'complete';
 
     const GUEST_GROUP                = 'NOT LOGGED IN';
@@ -216,6 +238,10 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
      */
     private $_cacheTypeList;
     /**
+     * @var \Magento\Framework\App\CacheInterface|null
+     */
+    private $_cache;
+    /**
      * @var \Magento\Customer\Model\ResourceModel\Attribute\CollectionFactory
      */
     private $_attCollection;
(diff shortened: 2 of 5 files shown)


dipola (approved):
Approving. The negative cache is right, the two admin clears are right, and the `else` branch is a case my issue named but my suggested fix would not have caught.

## I verified the no-op finding rather than take it, and it is exactly as you describe

This is the most valuable thing in the PR, so it deserved checking against the real source rather than the reasoning. From `vendor/magento/framework` on a 2.4 install:

```php
// Code/Reader/ClassReader.php:47 — what gets recorded as "required"
!$parameter->isOptional() && !$parameter->isDefaultValueAvailable(),

// ObjectManager/Factory/AbstractFactory.php:259-270
} elseif ($paramRequired) {
    $argument = ['instance' => $paramType];
    ...
} else {
    $argument = $paramDefault;
}
```

A default value makes `required` false; a not-required parameter takes the `else` and is handed `$paramDefault`. `['instance' => $paramType]` is only ever reached on the required branch, so the type is **never resolved**. Without the `di.xml` line `$cache` arrives `null` on every instantiation, `jsUrlLookupFailedRecently()` returns false forever, and the whole change is dead code.

The part worth dwelling on is how it fails: **silently, with a green suite**, because every test injects the double itself. That is the same shape as the vacuous assertion on PR 2356 yesterday and the library gate the week before — a check that passes with and without the defect. You found this one by going to a running install, which is the only thing that would have.

`testTheCacheIsOptionalInTheConstructorAndThereforeWiredInDi` is the right answer to it, and I checked it actually bites: the xpath targets `Ebizmarts\MailChimp\Helper\Data`, the argument name matches the constructor's `$cache` exactly, and it asserts the resolved type rather than merely that *something* is named. Delete either half — the default or the `di.xml` line — and it fails.

## The `else` is the case I would have missed

My issue named the 200-with-no-connected-site as one of the ways in, and then proposed a fix that only marked on throw. Nothing throws there, nothing is logged, nothing is written — so it would have kept asking forever and been the one path with no error to notice it by. Catching that in the `else` closes the issue properly rather than mostly.

Same for marking in the `catch` rather than adding a new one: `Mailchimp.php:288` raises `Mailchimp_HttpError` on `curl_error()` as well as on an HTTP status, so the ten-second timeout lands in the same handler as the 404. That is what makes the outage case — the widest one — actually covered.

## The clear-by-tag call is right, and for the reason you give

`MonkeyStore::beforeSave` can carry a website id or `0` while the marker is keyed by store view, so being surgical there would have been **wrong**, not merely more work. Dropping every marker costs at most one call per store view on the next render, which is what those store views were paying on every render before this. Correct trade, and correctly explained in the code rather than in the PR.

## One thing to carry forward, not a change request

The marker makes the failure **cheap** but leaves it **silent to the shopper**. Past the guard `$url` is empty, so the early return still emits `<script type="text/javascript" src="" defer="defer">` — which you confirmed while testing. So after this merges, that empty `src` is the last visible symptom of the whole family, and it is the one a merchant's own browser console shows them. Worth doing next; it does not belong in this PR.

## Measurements

4 uncached renders 4 calls → 1, and 1.87/1.14/1.24/1.31s → 1.16/0.39/0.41/0.35s. The cross-process figures are the ones that prove the marker survives a request rather than a request-scoped memo — 0.788s then 0.000s, and 0.642s again after `clearJsUrlFailures()`. That last step is what makes it a measurement of the clear path too, not just of the cache.

Noted for anyone testing this: the `di.xml` change needs a `cache:flush` before it takes effect, because merged DI config lives in the `config` cache type even in developer mode.

PR #2359, opened 2026-09-10
Load the interest-group tree only when a webhook row needs it

gonzaloebiz:
Found by reading telemetry from a live merchant, not from a bug report.

## The problem

`_loadGroups()` is the first statement of `processWebhooks()`, and the queue it exists to serve is read two lines later:

```php
public function processWebhooks()
{
    $this->_loadGroups();                                   // API work
    $collection = $this->_webhookCollection->create();
    $collection->addFieldToFilter('processed', ['eq' => self::NOT_PROCESSED]);
```

It walks every Mailchimp-enabled store view and issues **one `interest-categories` call plus one `interests` call per category**. On a `*/5` cron that is the whole tree, per store view, 288 times a day, whether or not anything will consume it.

The tree has exactly one consumer:

```
written   Cron/Webhook.php:361
read      Cron/Webhook.php:376   inside _getGroups()
          _getGroups()      <- _processMerges() :283
          _processMerges()  <- _profile() :239
          _profile()        <- case TYPE_PROFILE :119
```

So it is fetched to serve `profile`-type rows carrying a GROUPINGS merge field, and nothing else. `protected $groups = []` is a plain instance property — no cache, nothing survives the process.

On the merchant this was found on: 21 `lists` calls and 232 KB downloaded every five minutes, `bytes_up` 0, five consecutive beacons byte-identical. 40.1 of a 41.1-second cron run.

## The change

The load moves to `_getGroups()`, its only reader. A run with nothing to serve costs nothing; a run processing a hundred subscribe or unsubscribe rows also costs nothing, because those never reach the consumer.

**The guard is an explicit loaded-flag, not a check on the array.** An audience that defines no interest categories legitimately loads as `[]`, and `if (!$this->groups)` would read that as "not loaded yet" and re-fetch the whole tree for every row that asked — reintroducing the amplification, inside the guard meant to prevent it. The flag is set *before* the work rather than after, so a run that throws does not re-enter once per row.

The `isApiKeyFailed` guard inside the loop keeps its purpose but lost its rationale, which said this ran before any webhook work was looked at. It no longer does, so the comment now says what is true.

## Verified against the live API, not only in tests

2.4.8 install, empty queue, real Mailchimp credentials:

| `processWebhooks()`, empty queue | run 1 | run 2 |
|---|---|---|
| before | 1.576 s | 1.502 s |
| after | **0.005 s** | **0.001 s** |

And the half that matters more — that it did not become cheap by ceasing to work. Driving the consumer on the real cron object:

```
_getGroups() first call:  1.464s   <- loads the tree
_getGroups() second call: 0.000s   <- the flag holds
groups loaded: 6
groupsLoaded: true
```

Six real groups from the install's audience. The work moved to the point of use rather than disappearing.

Unit suite 183 → 187 tests, 362 assertions. The new tests were each checked against a mutant: restoring the eager load fails the empty-queue test; removing the trigger from `_getGroups()` fails three; **swapping the flag for a falsy check fails exactly the no-categories test**; removing the guard entirely fails two.

`phpcs --standard=Magento2`: 0 errors, and `Cron/Webhook.php` goes from 61 warnings to 59.

## Deliberately not changed

`:361` merges every store view's interests into one flat array with `array_merge_recursive`, and `_getGroups()` then matches on `name` + `category_id` across that merged set — so two store views on different audiences can collide on a category id. Making the load lazy neither causes nor worsens it. Keying the cache per store view would change matching behaviour for multi-view installs, which deserves its own change and its own review rather than riding along in this one.

## One thing worth knowing about the measurement it came from

The beacon cannot establish that the queue was empty on those runs. `_subscribe()`, `_unsubscribe()`, `_clean()` and `_updateEmail()` make no API calls at all, and neither does the `_processMerges()` branch of `_profile()` — which is the branch that consumes the tree. The only other call in the file is a `members->get` reached when a profile row matches no local customer *and* no local subscriber. So a run that processed a hundred rows would emit a byte-identical beacon.

That does not weaken the case for this change, since the fetch is unconditional either way. It does mean the saving is "the tree is no longer fetched when nothing needs it" rather than a claim about how many of those fetches were provably wasted.

### DIFF
--- a/Cron/Webhook.php
+++ b/Cron/Webhook.php
@@ -57,6 +57,18 @@ class Webhook
     protected $groups = [];
 
     /**
+     * Whether _loadGroups() has already run in this process.
+     *
+     * A flag rather than `if (!$this->groups)`: an audience that defines no
+     * interest categories legitimately loads as [], and a falsy check would
+     * read that as "not loaded yet" and re-fetch the whole tree for every
+     * webhook row that asked.
+     *
+     * @var bool
+     */
+    private $groupsLoaded = false;
+
+    /**
      * Webhook constructor.
      * @param \Ebizmarts\MailChimp\Helper\Data $helper
      * @param \Magento\Newsletter\Model\SubscriberFactory $subscriberFactory
@@ -87,7 +99,6 @@ class Webhook
     }
     public function processWebhooks()
     {
-        $this->_loadGroups();
         /**
          * @var $collection \Ebizmarts\MailChimp\Model\ResourceModel\MailChimpWebhookRequest\Collection
          */
@@ -336,15 +347,35 @@ class Webhook
         $subscriber->setIsStatusChanged(true);
         $subscriber->getResource()->save($subscriber);
     }
+    /**
+     * Fetch every enabled store view's interest-group tree, once per process.
+     *
+     * This used to be the first statement of processWebhooks(), which read the
+     * queue two lines later: an install with nothing to process paid the whole
+     * tree -- one `interest-categories` call per store view plus one
+     * `interests` call per category -- every five minutes, and discarded it.
+     * The single consumer is _getGroups(), reached only from a `profile`
+     * webhook carrying a GROUPINGS merge field, so it is now called from there.
+     *
+     * The flag is set before the work rather than after, so a run that throws
+     * does not re-enter this once per row.
+     *
+     * @return void
+     */
     protected function _loadGroups()
     {
+        if ($this->groupsLoaded) {
+            return;
+        }
+        $this->groupsLoaded = true;
+
         foreach ($this->storeManager->getStores() as $storeId => $val) {
             if (!$this->_helper->isMailChimpEnabled($storeId)) {
                 continue;
             }
-            // This runs before any webhook work is looked at, so an install
-            // with nothing to process still paid one rejected call per store
-            // view to find that out. One answer per credential is enough.
+            // One answer per credential is enough. This loop no longer runs
+            // unless a row actually needs the groups, but when it does run a
+            // dead key would still cost one rejected call per store view.
             if ($this->_helper->isApiKeyFailed($storeId)) {
                 continue;
             }
@@ -376,6 +407,8 @@ class Webhook
     }
     protected function _getGroups($groups, $cat)
     {
+        $this->_loadGroups();
+
         $rc = [];
         $gr = explode(",",$groups);
         foreach ($gr as $g) {
--- a/Test/Unit/Cron/WebhookLazyGroupsTest.php
+++ b/Test/Unit/Cron/WebhookLazyGroupsTest.php
@@ -1,0 +1,220 @@
+<?php
+/**
+ * Ebizmarts_MailChimp Magento Component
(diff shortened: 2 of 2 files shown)


dipola (approved):
Approving. The flag is right, the placement of the flag is right, and the mutation checks are what make the tests worth having.

## The two things that are easy to get wrong, both correct here

**A flag and not `if (!$this->groups)`.** An audience defining no interest categories loads as `[]`, and a falsy guard reads that as "not loaded yet" — so the store that pays most under the old code (nothing to fetch, fetched every time) would have kept paying, now once per webhook row, *inside* the guard added to stop it. That is the `=== null` lesson from the interest-cache work one file over, and it is good to see it applied rather than rediscovered.

**Set before the work, not after.** `_getGroups()` is called once per GROUPINGS item inside `_processMerges()`, so a throw with the flag set afterwards would re-enter the whole tree fetch per item — the amplification, moved rather than removed. Setting first makes a failed load cost one attempt per process.

Verified the consumer chain is still exactly one path: `$this->groups` written at `:361`, read only at `:376` inside `_getGroups()`, reached from `_processMerges` `:283` ← `_profile` `:239` ← `case TYPE_PROFILE` `:118`. Nothing else can reach it, so triggering from `_getGroups()` covers every reader.

## One consequence worth stating, not changing

`getApi()` sits **outside** the `try` (`:355`, with the try opening at `:356`), so a store view whose API construction throws takes `_loadGroups()` out with it. With the flag already set that leaves `$this->groups` half-built and treated as complete for the rest of the process — the later store views never load, and `_getGroups()` matches against a partial tree without knowing it.

That is still the right trade: re-entering would restore the per-row amplification, and the cron runs again in five minutes. But it is a real behaviour, it is invisible when it happens, and it is the same shape as the `getInterest()` latent defect — one `try` around a `1 + N` loop, first failure abandons the rest and returns what it has. One line in the docblock saying a partial load is deliberately kept would stop the next reader having to work out whether it was considered.

## The mutants are the part that makes this evidence

> restoring the eager load fails the empty-queue test, removing the trigger fails three, **swapping the flag for a falsy check fails exactly the no-categories test**, removing the guard fails two.

That third one is the whole point. `testAnAudienceWithNoCategoriesIsNotRefetchedPerRow` would pass against a falsy-guarded implementation for every audience that *has* categories, which is most of them — checking it against the mutant is what proves it discriminates. Four tests, four distinct mutants, each failing a different subset.

## On the per-view keying you dropped

Right call, and right to say so plainly rather than let me notice the absence. `:361` merges every view's interests with `array_merge_recursive` and `_getGroups()` matches on `name` + `category_id` across the merged set, so keying per view changes *matching* behaviour on multi-view installs. That is a semantics change wearing a performance fix's clothes, and it belongs in its own review. I will file it separately.

## And the claim you narrowed

Agreed, and it is the honest form: what the change earns is "the tree is no longer fetched when nothing needs it" — four of the five handlers make no API calls at all. Your 1.576 s → 0.005 s on one view with a small audience is that effect at the scale you can measure directly, and it is the right evidence to cite.

PR #2360, opened 2026-09-10
Do not emit a script tag when the connected-site URL is empty

gonzaloebiz:
One line of behaviour, filed as cosmetic and promoted by a review note: after the negative-cache work, this empty tag is the **last visible symptom** of that whole family, and the only one a merchant sees in their own browser console.

## What it does

`view/frontend/templates/mailchimpjs.phtml` rendered the script tag unconditionally. An unresolved connected site leaves the URL empty — `null` rather than `''`, since the config read returns null when nothing is saved — so the page carried:

```html
<script type="text/javascript" src="" defer="defer"></script>
```

That is inert: per the HTML spec an empty `src` fires an `error` event and fetches nothing. It is **not** why the connected-site script fails to load — it is a symptom of the URL being empty, which is addressed at the source. But it is the symptom a merchant can see, so it invites an investigation into the wrong thing. Rendering nothing says as much and asks no questions.

## Verified on a running install, both directions

Because the risk in a guard like this is not the empty case, it is silencing the working one.

| store id | before | after |
|---|---|---|
| does not resolve | `<script type="text/javascript" src="" defer="defer">` | **no tag at all** |
| resolves | tag renders | **tag renders, unchanged** |

The "before" row is a real capture from this morning against the untouched template, not a reconstruction. The healthy row was re-checked after a late whitespace fix, so what was measured is what is in the commit.

Unit suite unchanged at 183 tests, 357 assertions. `phpcs --standard=Magento2`: 0 errors and **no new warnings** — the first version of the guard added two, from `if ($url) :` spacing and the unindented body, and both are fixed here.

## Not in scope

The pixel block emits its own `mc-pixel-config` JSON carrying a script URL from a different config path. It is untouched by this and was confirmed present in the same renders, so nothing here changes the pixel.

### DIFF
--- a/view/frontend/templates/mailchimpjs.phtml
+++ b/view/frontend/templates/mailchimpjs.phtml
@@ -6,5 +6,13 @@
 $url = $block->getJsUrl();
 $render = $block->getRender();
 ?>
-<?= /* @noEscape */ $render->renderTag('script',['type'=>'text/javascript','src'=>$escaper->escapeUrl($url),'defer'=>'defer'],"\n",true); ?>
-
+<?php // An unresolved connected site leaves $url empty -- null when nothing is
+      // saved, since the config read returns null rather than ''. Rendering
+      // anyway put <script src=""> on the page: inert per the HTML spec, which
+      // fires an error event and fetches nothing, but it is the one symptom of
+      // that state a merchant sees in their own browser console, and it is not
+      // the reason the connected-site script is missing. No tag says the same
+      // thing without inviting the wrong investigation. ?>
+<?php if ($url): ?>
+    <?= /* @noEscape */ $render->renderTag('script',['type'=>'text/javascript','src'=>$escaper->escapeUrl($url),'defer'=>'defer'],"\n",true); ?>
+<?php endif; ?>


dipola (approved):
Approving. Ten lines, and it does slightly more than either the issue or the PR claims.

## The truthiness point is right, and I checked why

`Block\Mailchimpjs::getJsUrl()` reads the value straight off `ScopeConfigInterface::getValue()` (103.4.81, `Block/Mailchimpjs.php:53`), and that returns **null** for an unset path — not `''`. An `$url !== ''` guard would have passed on null and rendered the tag anyway, i.e. changed nothing in the exact state this exists for. `if ($url)` is the correct shape.

## It also covers a case nobody has been counting

Reading the whole block method rather than the diff:

```php
if ($active && !$url) {
    $url = $this->_helper->getJsUrl($storeId);
}
return $url;
```

The helper is only consulted when the integration is **active**. So a store with Mailchimp switched off never reaches the lookup at all, `$url` stays null, and before this PR that store emitted `<script type="text/javascript" src="" defer="defer">` on **every page of every render**, forever, with no failure anywhere to explain it.

That is a larger population than the failing stores this was filed for, and it is the one where the empty tag is most misleading — there is no fault to find, because nothing is wrong. Worth knowing when someone asks why a one-line template change was worth a release slot.

## On the reclassification

You are right that my note on 2358 moved this, and right about the reason. Once the negative cache lands, the network cost is gone and the empty `src` is the only part of that whole family a merchant can still see — in their own console, with no error next to it, pointing at the wrong thing. Cheap to fix, and it stops the next person spending an afternoon on a tag that was never the cause.

## Small things, none blocking

The `escapeUrl` and the `@noEscape` annotation are both preserved, so this changes what is rendered and not how. And when `$url` is empty the template now emits a couple of newlines instead of the tag, which is inert and not worth an `ob_start`.

Nothing else. Merge it.

PR #2361, opened 2026-09-10
Reject the store dropdown's non-answers, and keep a good list id on a failed lookup

gonzaloebiz:
Two defects in `Model/Config/Backend/MonkeyStore::beforeSave`, both on the admin save path.

## 1. It persisted values that are not stores

The dropdown offers two: **-1**, its placeholder, and **0**, the `---No Data---` option it falls back to when there is no API key at that scope or the store listing threw (`Model/Config/Source/MonkeyStore.php`). `beforeSave` stored either as though a store had been chosen.

The read side already treats both as inert, so this no longer arms anything at render time. What is left is a setting holding a value that is not a store, which every later reader has to know to special-case — and three already do (`Cron/Ecommerce.php`, `Cron/GenerateStatistics.php`, `Controller/Adminhtml/Ecommerce/Getaccountdetails.php`).

**The rejection is scoped to ecommerce being active, and that scoping is the whole care of this change.** With ecommerce off, "no store selected" is a legitimate state; rejecting there would leave a merchant unable to save *any* Mailchimp configuration — including the switch that would turn ecommerce off. The guard sits under the same `$active` the method already computes.

## 2. A failed lookup overwrote the configured audience with nothing

When the store value changed, `getStore()` answers `null` if its own lookup fails — a 404 on a store Mailchimp no longer has, a revoked key — and that `null` went straight into `XML_PATH_LIST` unconditionally.

So a failed lookup did not merely fail: **it destroyed the answer it could not confirm**, which is a plausible route to a configured audience that resolves to nothing.

The same `null` then reached `createWebHook($apiKey, null)`, registering a webhook against no audience. Both are guarded now, and what is already configured survives a lookup that could not confirm it.

## Verified

**Against the running install, through the real object graph** — the backend model built by the ObjectManager, not by hand:

```
ecommerce ON,  value=-1   -> REJECTED: Select a Mailchimp store, or turn Mailchimp ecommerce off...
ecommerce ON,  value=0    -> REJECTED
ecommerce ON,  value=""   -> REJECTED
ecommerce OFF, value=-1   -> ACCEPTED
```

And the premise of the second defect, confirmed against the live API on that install rather than assumed:

```
getStore() with a store id that 404s -> NULL
```

That is the value that used to be written over the audience id.

**Unit suite 183 → 197 tests, 372 assertions.** Each behaviour was checked against a mutant:

| mutation | result |
|---|---|
| drop the rejection | 5 failures |
| reject with ecommerce off too | 5 errors |
| `isARealStore` checks only -1 | 3 failures |
| write the null list id again | 1 failure |
| create the webhook on a null list again | 1 failure |

`phpcs --standard=Magento2`: 0 errors, and `MonkeyStore.php` stays at the same 11 warnings it had.

## What was not exercised end to end

The rejection path returns early and writes nothing, so it was safe to run for real. The **null-audience path was not** — running it on the install would delete connected-site URLs, cancel pending batches and reset errors for a store view, which is more than a verification should cost on a shared box. That half rests on the unit tests and their mutants, plus the live confirmation that `getStore()` really does answer null. Saying so rather than letting the section above imply otherwise.

### DIFF
--- a/Model/Config/Backend/MonkeyStore.php
+++ b/Model/Config/Backend/MonkeyStore.php
@@ -21,6 +21,7 @@ use Magento\Framework\App\Config\ScopeConfigInterface;
 use Magento\Framework\App\Config\Value;
 use Magento\Framework\Data\Collection\AbstractDb;
 use Magento\Framework\Model\Context;
+use Magento\Framework\Exception\LocalizedException;
 use Magento\Framework\Model\ResourceModel\AbstractResource;
 use Magento\Framework\Registry;
 use Magento\Framework\Stdlib\DateTime\DateTime;
@@ -73,6 +74,24 @@ class MonkeyStore extends Value
         } else {
             $active = 0;
         }
+        // The dropdown offers two values that are not stores: -1, its
+        // placeholder (Model/Config/Source/MonkeyStore.php), and 0, the
+        // '---No Data---' option it falls back to when there is no API key at
+        // this scope or the store listing threw. Persisting either leaves a
+        // setting that is not a store, which every later reader has to know to
+        // special-case, and three already do.
+        //
+        // Only while ecommerce is active. With it off, "no store selected" is a
+        // legitimate state, and rejecting it here would stop a merchant saving
+        // any Mailchimp configuration at all -- including the switch that would
+        // turn ecommerce off.
+        if ($active && !$this->isARealStore($this->getValue())) {
+            throw new LocalizedException(
+                __('Select a Mailchimp store, or turn Mailchimp ecommerce off. '
+                    . 'Ecommerce cannot sync without one.')
+            );
+        }
+
         if ($active && $this->isValueChanged()) {
             $mailchimpStore     = $this->getOldValue();
             // charge the $newListId
@@ -85,12 +104,20 @@ class MonkeyStore extends Value
                 $newListId = $data['general']['fields']['monkeylist']['value'];
             } else {
                 $newListId = $this->getStore($apiKey, $this->getValue());
-                $this->_helper->saveConfigValue(
-                    Data::XML_PATH_LIST,
-                    $newListId,
-                    $this->getScopeId(),
-                    $this->getScope()
-                );
+                // getStore() answers null when its own lookup fails -- a 404 on
+                // a store Mailchimp no longer has, a revoked key. Writing that
+                // replaced a working audience id with nothing, so a failed
+                // lookup did not merely fail: it destroyed the answer it could
+                // not confirm. Keep what is there and let the failure be a
+                // failure.
+                if ($newListId) {
+                    $this->_helper->saveConfigValue(
+                        Data::XML_PATH_LIST,
+                        $newListId,
+                        $this->getScopeId(),
+                        $this->getScope()
+                    );
+                }
             }
             $this->oldListId = $this->_helper->getConfigValue(
                 Data::XML_PATH_LIST,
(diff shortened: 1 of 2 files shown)


dipola (approved):
Approving — both fixes are right, and the second one is the more valuable of the two. One thing I would like you to look at before merging, and one confirmation.

## The framing on the list id is the finding

> a failed lookup did not merely fail, **it destroyed the answer it could not confirm**

That is a better description than the one I filed, and it is the reason the third consequence exists at all. Guarding the webhook registration with the same `$newListId` is the right catch — `createWebHook($apiKey, null)` registering against no audience is a worse outcome than the overwrite, and I had not followed it that far.

## `isARealStore()` — checked, and it holds on both engines

`(bool)$value && $value != -1`:

| value | result | |
|---|---|---|
| `0`, `'0'`, `''` | rejected | falsy |
| `-1`, `'-1'` | rejected | loose compare, and the form submits a string |
| any 32-character hex store id | accepted | PHP 8 casts the **int to string** for a non-numeric string comparison, so the comparison against `'-1'` is true |

Worth noting the PHP 8 direction explicitly because it is the opposite of PHP 7's, where the string would have been cast to `2`. Both reach the same answer here, so the loose compare is safe on either — but it is safe for different reasons, which is the kind of thing that decays quietly.

## The thing I would like you to check: `$active` is truthy whenever the field is INHERITED

Not introduced by this PR, but this PR is what makes it bite. The existing ladder:

```php
if (isset($data['ecommerce']['fields']['active']['value'])) {
    $active = $data['ecommerce']['fields']['active']['value'];
} elseif ($data['ecommerce']['fields']['active']['inherit']) {
    $active = $data['ecommerce']['fields']['active']['inherit'];   // <- the inherit FLAG, not the inherited VALUE
}
```

At a website or store-view scope with "Use Default" ticked, Magento posts `inherit => 1` and **no** `value`, so `$active` becomes `1` — regardless of whether the inherited setting is on or off.

So a scope that inherits **ecommerce = off** and holds `-1` now gets the rejection, and the message tells them to *"turn Mailchimp ecommerce off"* when it already is. The escape exists (untick inherit, set 0 explicitly at that scope) but nothing on screen points at it, and it is the lockout shape you deliberately designed against in the direct case.

Two ways out, your call: resolve the inherited value rather than reading the flag, or leave it and say so in the comment so the next reader knows it was weighed.

## And one confirmation, because it decides who meets this first

The rejection sits above `if ($active && $this->isValueChanged())`, so it fires on **every** save of that section while ecommerce is active — not only when the store value changed. A merchant sitting on `-1` today therefore hits it the next time they save anything in Mailchimp config, whether or not they touched the dropdown.

I think that is right, and it is certainly the behaviour that gets the value out of circulation fastest. Worth being a decision rather than a side effect of placement.

Neither of those is blocking. The tests cover the two rejection paths, both scopes and both null-guard consequences, and the parameterised value list is the right shape for a guard whose whole content is which values count.

gonzaloebiz:
Both points answered, one with a commit.

## The inherit flag — fixed, `1aefbcae`

You were right and it is worse than "not introduced here": harmless before, because nothing gated on
`$active` in a way a merchant could hit. This PR is what gives it teeth, so it belongs in this PR.

A scope inheriting ecommerce **off** and holding `-1` would have been rejected and told to turn
ecommerce off — with no field at that scope it could change to comply, because the one it would have to
change is the one it is inheriting. That is precisely the lockout the guard was scoped to avoid,
arriving through the back door.

I resolved the inherited value rather than documenting the flag. Documenting would have left a message
that lies to the one merchant who reaches it.

Two tests, and the mutant confirms they bite: putting the flag back makes
`testAScopeInheritingEcommerceOffAcceptsTheNonStore` fail, while the inheriting-**on** case still
rejects.

## Firing on every save — deliberate, and now written down

Your reading is the one I intended, and you are right that it should be a decision rather than a
consequence of where the lines sit. It is above the `isValueChanged()` gate on purpose: an install
already holding `-1` meets it on its **next save of that section**, not only if someone happens to
reopen the dropdown. That is what takes the value out of circulation instead of preserving it.

Given your measurement — `-1` across 11 store paths on 4 installs — the alternative is those 11 keeping
the value indefinitely, since nothing else would ever prompt a change. The comment now says this, so it
does not read as an accident of placement.

## On `isARealStore()` and the PHP 8 direction

Noted, and it is the more useful half of that check: the loose compare is safe on both engines **for
different reasons**, which is exactly the kind of thing that decays without anyone noticing. The
parameterised value list is what would catch it if it ever stopped being safe — every value the guard
claims to handle is in the provider, so an engine change that flips one shows up as a named failure
rather than as behaviour drift.

Suite 187 → 199 tests, 374 assertions. `phpcs`: 0 errors, file unchanged at 11 warnings.

PR #2362, opened 2026-09-10
Say that a partial interest-group load is deliberate

gonzaloebiz:
Comment only. No behaviour changes.

Raised in review of the lazy-load change, and left out of that PR rather than pushed onto it after approval.

## The behaviour

`getApi()` sits outside the `try` in `_loadGroups()`, so a store view whose API construction throws takes the whole loop with it. The loaded flag is set before the work, so `$this->groups` stays half-built and is treated as complete for the rest of the process: later store views never load, and `_getGroups()` matches against a partial tree without knowing it.

Nothing reports this. It is invisible when it happens.

## Why it stays

Re-entering on failure would restore exactly the per-row amplification the method was changed to remove, and the job runs again in five minutes — so a partial load costs one cycle where a retry loop would cost a call per webhook row.

That is a considered trade, and nothing said so. The next reader would have had to work out from scratch whether it had been thought about, which is the cost this line removes.

## Note for whoever reads this next

It is the same shape as the `getInterest()` behaviour noted on the storefront-latency issue: one `try` around a `1 + N` loop, where the first failure abandons the rest and returns what it has. Two instances now, in two files. If a third turns up it is a rule worth fixing as one, rather than three call sites.

Suite unchanged at 187 tests, 362 assertions. `phpcs --standard=Magento2`: 0 errors, and the file goes from 60 warnings to 59.

### DIFF
--- a/Cron/Webhook.php
+++ b/Cron/Webhook.php
@@ -358,7 +358,18 @@ class Webhook
      * webhook carrying a GROUPINGS merge field, so it is now called from there.
      *
      * The flag is set before the work rather than after, so a run that throws
-     * does not re-enter this once per row.
+     * does not re-enter this once per row. That has a consequence worth stating
+     * because it is invisible when it happens: getApi() is outside the try
+     * below, so a store view whose API construction throws takes this whole
+     * loop with it, and the flag is already set -- leaving $this->groups
+     * half-built and treated as complete for the rest of the process. Later
+     * store views never load, and _getGroups() matches against a partial tree
+     * without knowing it.
+     *
+     * That is deliberate rather than overlooked. Re-entering on failure would
+     * restore exactly the per-row amplification this method was changed to
+     * remove, and the job runs again in five minutes, so a partial load costs
+     * one cycle where a retry loop would cost a call per webhook row.
      *
      * @return void
      */


dipola (approved):
Approving. This is the right response to that note — the behaviour was already correct, what was missing was that a reader could tell it had been chosen.

Two things it gets right that a weaker version of this comment would not:

**It states the mechanism, not just the conclusion.** "`getApi()` is outside the try below" is the fact that makes the rest inevitable, and it is the one a reader cannot see from this method alone — the `try` is twenty lines down and opens after the call that can throw. Someone who only reads "a partial load is deliberate" has to take it on trust; someone who reads this can check it.

**It says the failure is invisible.** That is the part that decides whether the trade is acceptable, and it is the part that usually goes unwritten. A half-built tree that `_getGroups()` matches against without knowing is not a degraded result the caller can detect — it is a wrong answer wearing a correct one's shape. Naming that is what makes "deliberate" mean something rather than being a label.

The trade itself I already agreed with and still do: re-entering restores the per-row amplification the whole change existed to remove, and one cycle of a five-minute job is the cheap side of that. Worth noting the asymmetry explicitly, since the comment implies it — the cost of the partial load is bounded by the cron interval, while the cost of retrying is unbounded in the number of webhook rows, which is the variable that made this a problem in the first place.

Nothing to change. Comment only.

dipola:
All five of these are merged and reviewed, and nothing of ours is open behind them — **please cut 103.4.83.**

```
#2358  #2359  #2360  #2361  #2362
```

**One difference from #2357, so it does not catch you out:** last time `composer.json` and `etc/module.xml` already carried the new number, because the bump had arrived early with an unrelated PR. They do not this time — both still read **103.4.82** on `develop-2.4`. So 83 needs the version bump as well as the changelog, i.e. the full shape rather than the changelog-only one.

The only other open PR here is #1757 from an outside contributor, which has never been reviewed and is not ours to hold a release for.

Two of these five are the reason I would not let it sit until next week:

**#2358** is the one a shopper feels. While Mailchimp is slow or unreachable, an unresolved connected-site lookup blocks the page render for up to the library's ten-second ceiling — on every uncached render, cart and checkout included. Measured: 1.87s → 1.16s on a first render, 1.24s → 0.41s on a repeat. Until it is tagged, every store in that state is still paying it per visitor.

**#2361** is the one that quietly loses data. A failed lookup was overwriting a correct audience id with nothing and then registering a webhook against nothing — a check that did not merely fail, but destroyed the answer it could not confirm. That keeps happening on every affected save until this ships.

The other three are smaller but real: a five-minute cron that fetched an interest tree it discarded, an empty script tag on every page of any store with the integration switched off, and a comment recording why a partial load is deliberate.

PR #2363, opened 2026-09-10
version 83 for magento 2.4

gonzaloebiz:
> [!IMPORTANT]
> **Merge this after #2364, not before.**
>
> #2364 moves the library floor to `>=3.0.49` together with the `MIN_LIB_VERSION` constant that has to
> agree with it. This PR deliberately does not touch either — they belong in one place, and having them
> in two is how they drifted apart in the first place.
>
> Merged in the wrong order, 103.4.83 would be tagged asking for `>=3.0.47` while warning against
> `3.0.45`: the inconsistency published in a tag, which is the form that is hardest to correct
> afterwards. This branch will be refreshed from `develop-2.4` before it is merged, so the diff shows
> exactly what the tag will contain.

Version bump, changelog, and the library floor. No behaviour changes of its own.

## What it releases

Five PRs merged to `develop-2.4` since 103.4.82 was cut earlier today. All five are in milestone 10x.x.83.

**Fewer calls to the Mailchimp API from merchant servers**

- **2358** — a connected-site lookup that fails is remembered, instead of repeating on every uncached render. Measured on a 2.4.8 install: 4 uncached renders went from 4 API calls to 1.
- **2359** — the interest-group tree is no longer fetched at the top of the webhook cron, which read its queue two lines later. On the merchant this was found on, that was 21 calls and 232 KB every five minutes, discarded.

**Fixed bugs**

- **2361** — the store dropdown's `-1` and `0` are no longer persisted as though they were stores, and a failed audience lookup no longer overwrites a working list id with null, nor registers a webhook against no audience.
- **2360** — no `<script src="">` when the connected-site URL is empty.

**Merged pull requests**

- **2362** — records that a partial interest-group load is deliberate.

## Two things this PR carries that the last release PR did not

**The bump itself.** `composer.json` and `etc/module.xml` still said 103.4.82. For 103.4.82 the bump had arrived earlier with a feature PR, so that release PR only touched the changelog. This one does all three.

**The library floor moves to `>=3.0.49`.** That is the release that turns contact sharing on wherever nobody has declined. It stays a floor rather than a pin, as it has been since 3.0.44, so an install already resolving to something newer is unaffected. 3.0.49 is published on Packagist and on the mirror.

Suite 203 tests, 379 assertions. `composer validate`: valid.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,30 @@
 # Changelog
 
+## [103.4.83](https://github.com/mailchimp/mc-magento2/tree/103.4.83)
+
+[Full Changelog](https://github.com/mailchimp/mc-magento2/compare/103.4.82...103.4.83)
+
+**Fewer calls to the Mailchimp API from merchant servers:**
+
+- Remember a failed connected-site lookup instead of repeating it every render [\#2358](https://github.com/mailchimp/mc-magento2/pull/2358)
+- Load the interest-group tree only when a webhook row needs it [\#2359](https://github.com/mailchimp/mc-magento2/pull/2359)
+
+**Fixed bugs:**
+
+- Reject the store dropdown's non-answers, and keep a good list id on a failed lookup [\#2361](https://github.com/mailchimp/mc-magento2/pull/2361)
+- Do not emit a script tag when the connected-site URL is empty [\#2360](https://github.com/mailchimp/mc-magento2/pull/2360)
+
+**Merged pull requests:**
+
+- Say that a partial interest-group load is deliberate [\#2362](https://github.com/mailchimp/mc-magento2/pull/2362)
+
+**Dependencies:**
+
+The library floor moves to `>=3.0.49`, which is the release that turns contact
+sharing on wherever nobody has declined. The constraint is a floor rather than a
+pin, as it has been since 3.0.44, so an install already resolving to something
+newer is unaffected.
+
 ## [103.4.82](https://github.com/mailchimp/mc-magento2/tree/103.4.82)
 
 [Full Changelog](https://github.com/mailchimp/mc-magento2/compare/103.4.81...103.4.82)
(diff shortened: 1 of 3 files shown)


dipola (approved):
Approving, and the ordering note at the top is the right call — merge #2364 first.

Checked the contents against `develop-2.4` rather than the description:

- **Version bump present this time.** `composer.json` and `etc/module.xml` both move 103.4.82 → 103.4.83. That was the one difference from #2357 I flagged, where they had already been bumped early by an unrelated PR and the release was changelog-only. This one needed the full shape and has it.
- **Six PRs in the entry** — 2358, 2359, 2360, 2361, 2362, 2364 — against five merged when I asked for the cut, plus 2364 which lands before it. Nothing merged to `develop-2.4` since 103.4.82 is missing from it.

## On holding it for #2364

Agreed, and your reason is the one that matters: tagged in the wrong order, 103.4.83 would ship asking for `>=3.0.47` while warning against `3.0.45`, and an inconsistency inside a tag is the form that cannot be quietly corrected — it is what a merchant installs.

Refreshing the branch from `develop-2.4` before merging is the right follow-through, since otherwise the diff shown here is not the diff that gets tagged.

## What this actually delivers

Two of the six are the reason it was worth cutting today rather than next week:

**2358** — a shopper's page render no longer blocks up to the library's ten-second ceiling on every uncached render while Mailchimp is unreachable. Until this is tagged, every store in that state pays it per visitor.

**2361** — a failed lookup stops overwriting a correct audience id with nothing and registering a webhook against nothing. That one was destroying configuration on every affected save.

The other four are smaller and real: a five-minute cron that fetched an interest tree it discarded, an empty script tag on every page of any store with the integration switched off, a comment recording why a partial load is deliberate, and now the version comparison.

Tag it and it starts reaching merchants — 103.4.82 went to Packagist at 10:53 today and had a real install by 13:30, so the distribution side is not the slow part.

PR #2364, opened 2026-09-10
Compare the library version as a version, and align the floor it checks

gonzaloebiz:
Found in review of the QoS beacon work. Two defects that only look like one.

## 1. The comparison compares strings

`Block/Adminhtml/System/Config/Fieldset/Hint::checkLibVersion()` used a relational operator on version strings:

```php
if (Data::MIN_LIB_VERSION > $version) {
```

Measured, not reasoned — the same two values through both comparisons:

| installed | `MIN > $version` (string) | `version_compare($version, MIN, '<')` |
|---|---|---|
| `3.0.9` | false → **silent** | true → warns |
| `3.0.44` | true → warns | true → warns |
| `3.0.48.0` | false → silent | true → warns |
| `3.0.49` | false → silent | false → silent |
| `3.0.100` | true → **warns wrongly** | false → silent |

So the store nine releases behind was told nothing, and from `3.0.100` onwards everybody would have been warned. The warning was blind in exactly the direction it exists for.

`3.0.48.0` is in the table because that is the shape a real install reports — `InstalledVersions::getVersion()` returns a normalised four-segment string, which I checked on a running 2.4.8 rather than assuming three.

## 2. The constant had drifted behind the floor

`MIN_LIB_VERSION` sat at `3.0.45` while composer's floor moved twice without it.

They answer for **different populations, and only one of them is enforced**: composer holds the floor for installations installed with composer, while this constant is the only thing that says anything at all to an **app/code installation**, where the merchant runs whichever library happens to be present. A constant left behind the floor is silence aimed at exactly the installations that cannot be told any other way.

Both now say `>=3.0.49`, and `testTheConstantTracksTheComposerFloor` reads `composer.json` and asserts they agree, so they cannot drift apart silently again.

## Why the floor moves here and not in the release PR

It was in the release PR when I opened it. It belongs here: the floor and the constant express the same decision to two different audiences, and splitting them across two PRs is how they got out of step in the first place. The release PR now carries only the version bump and the changelog.

## Verified

11 tests. Both defects checked against mutants: restoring the string comparison fails three, and putting the constant back at `3.0.45` fails three.

Suite 203 → 214 tests, 390 assertions.

### DIFF
--- a/Block/Adminhtml/System/Config/Fieldset/Hint.php
+++ b/Block/Adminhtml/System/Config/Fieldset/Hint.php
@@ -72,7 +72,13 @@ class Hint extends \Magento\Backend\Block\Template implements
     public function checkLibVersion()
     {
         $version = $this->_moduleVersion->getLibVersion('ebizmarts/mailchimp-lib');
-        if (\Ebizmarts\MailChimp\Helper\Data::MIN_LIB_VERSION > $version) {
+        // version_compare, not `>`. Comparing version strings with a relational
+        // operator compares them as strings, so '3.0.45' > '3.0.9' is false and
+        // an installation running 3.0.9 was told nothing -- the oldest
+        // libraries, which are the whole reason this warning exists, were the
+        // ones it could not see. It fails in the other direction too: it would
+        // have warned every installation from 3.0.100 onwards.
+        if (version_compare((string)$version, \Ebizmarts\MailChimp\Helper\Data::MIN_LIB_VERSION, '<')) {
             return $version;
         } else {
             return false;
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -155,7 +155,20 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
 
     const MAX_MERGEFIELDS = 100;
 
-    const MIN_LIB_VERSION = '3.0.45';
+    /**
+     * The oldest library this extension is willing to run against without
+     * saying so in the admin.
+     *
+     * This must track the `ebizmarts/mailchimp-lib` floor in composer.json.
+     * They answer for two different populations and only one of them is
+     * enforced: composer holds the floor for installations that were installed
+     * with composer, and this constant is the only thing that says anything at
+     * all to an app/code installation, where the merchant runs whichever
+     * library happens to be present. A constant left behind the floor is
+     * silence aimed at exactly the installations that cannot be told any other
+     * way.
+     */
+    const MIN_LIB_VERSION = '3.0.49';
 
     protected $counters = [];
 
--- a/Test/Unit/Block/Adminhtml/LibVersionWarningTest.php
+++ b/Test/Unit/Block/Adminhtml/LibVersionWarningTest.php
@@ -1,0 +1,127 @@
+<?php
+/**
+ * Ebizmarts_MailChimp Magento Component
+ *
+ * @category    Ebizmarts
+ * @package     Ebizmarts_MailChimp
+ * @author      Ebizmarts Team <info@ebizmarts.com>
+ * @copyright   Ebizmarts (http://ebizmarts.com)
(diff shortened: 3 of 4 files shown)


dipola (approved):
Approving. The measured table is what makes this reviewable rather than arguable, and the alignment test is the part that stops it recurring.

## I checked whether there is a second one. There is not.

`version_compare` appears **nowhere** in the extension at 103.4.81, and `Hint.php:75` is the **only** relational comparison anywhere in the codebase with a version on either side. So this is not a first pass at a class of bug — it is the whole class, and this PR also introduces the first correct use of the function.

Worth saying because "we fixed one instance" and "there was one instance" are very different states to leave a reviewer in, and only the grep distinguishes them.

## The direction of the failure is the finding

`'3.0.45' > '3.0.9'` is false, so the installations nine releases behind — the entire reason the warning exists — were the ones it could not see, while everything from `3.0.100` on would have been warned wrongly. A check that is blind in exactly its own direction is worse than no check: it reads as evidence of health.

`3.0.48.0` in the table earns its place too. Four segments is what `InstalledVersions::getVersion()` actually returns, and you confirmed it on a running 2.4.8 rather than assuming three — which is the difference between the table being a demonstration and being an illustration.

## The alignment test is the structural half

Reading `composer.json` and asserting `'>=' . MIN_LIB_VERSION === $floor` is the right shape. "Remember to move both" is not a fix; a test that fails when they disagree is. And the docblock says *why* they are two things — composer enforces the floor for composer installs, the constant is the only thing an app/code install is ever told — which is the fact that makes the duplication necessary rather than sloppy.

## One question, not a blocker

The floor moves to `>=3.0.49`, and **nothing in 103.4.83 requires 3.0.49's API.** The library release is same-day, and the constant only needs to be ≤ the floor, not equal to the newest tag.

So this is policy rather than necessity: it pulls every composer install onto today's library on their next update, and it means a merchant pinned to 3.0.48 for any reason cannot install 103.4.83 at all. Both are probably what you want — the whole point is to stop the constant lagging — but it is a stronger constraint than the code needs, and the test now makes the floor and the constant move together forever, so the next person raising one raises the other.

If it is deliberate, ignore me. I would rather ask than find out from a merchant who pinned.

gonzaloebiz:
Deliberate, and worth recording as a decision rather than leaving it to be inferred.

The floor moving to `>=3.0.49` is policy, not necessity — you are right that nothing in 103.4.83 touches
3.0.49's API. It was an explicit instruction: the extension should ask for the current library. 3.0.49
is the release that turns contact sharing on wherever nobody has declined, and for **composer**
installations raising the floor is the only mechanism that carries it — the library reaches them on
their next update whether or not they touch the extension, which is precisely how the affected
population picked up 3.0.48 within six days of it shipping.

Your two consequences are both real and both accepted:

**A merchant pinned to 3.0.48 cannot install 103.4.83.** That is the cost. It is reversible in a later
release if anyone turns out to be in that position, and the alternative — leaving the floor where it was
— means the composer population does not get the library until something else happens to move it.

**The floor and the constant now move together forever.** That is the intent rather than a side effect.
They drifted apart precisely because nothing tied them, and the next person raising one raising the
other is the outcome the test exists to force.

Worth adding, since it changes who this actually reaches: the floor does **nothing** for an app/code
installation. I checked on a running 2.4.8 where the extension sits in `app/code` — the library is in
`vendor/` only because the **root** `composer.json` asks for it (`^3.0.47`), and `composer why` names
the project, not the module. Composer never reads the module's own requires there. So for that
population the constant is not a duplicate of the floor at all; it is the only statement that exists,
which is the case your §4.4 on the beacon issue already made from the fleet side.

Your grep is the more useful half of this review, and I should have run it: knowing `version_compare`
appears nowhere else, and that `Hint.php:75` was the only version comparison in the codebase, is what
turns "we fixed one" into "there was one". Recorded here so nobody re-audits for a second instance.

PR #2365, opened 2026-09-11
Ask for the audience's total contacts on the call already being made

gonzaloebiz:
One argument. The extension half of the per-audience member counts the QoS beacon has been carrying a column for since it was built and has never had a value for.

## What and why

`GenerateStatistics` reads `lists/{id}` every twelve hours. That response already carries the audience's member, unsubscribe and cleaned counts, and will also carry `total_contacts` if asked — **same call, same cadence, no quota, no latency**.

`total_contacts` is the one figure the payload cannot be made to yield by arithmetic. `member_count + unsubscribe_count` omits non-subscribed contacts entirely, and on an ecommerce account those are most of the audience — **413 against 20,971** on one account measured in the estate. The sum is not an approximation of the billable figure; it is a different number.

## Safe against an older library

The parameter is positional because that is the library's signature, and it lands in the library release after 103.4.83's floor. That is fine: **PHP ignores surplus arguments to a userland function**, so against a library without the parameter the field is simply absent rather than an error.

Verified both ways on a running 2.4.8:

| library on the box | result |
|---|---|
| **3.0.49** (no such parameter) | cron completes normally in 3.9 s, `total_contacts` absent |
| the library branch that has it | cron completes, and the beacon's bucket carries `list_total_contacts` |

The second run, end to end through the real cron rather than a synthetic call:

```
list_id                  '…'
list_member_count        3
list_unsubscribe_count   0
list_cleaned_count       0
list_total_contacts      5
```

## Nothing here reads it

This extension does not consume the new field, and the statistics payload is unchanged. It exists for the QoS beacon, which observes the response passively and can only report what a caller asked for — so without this argument, the library half delivers nothing.

### DIFF
--- a/Cron/GenerateStatistics.php
+++ b/Cron/GenerateStatistics.php
@@ -192,7 +192,30 @@ class GenerateStatistics
             if ($storeId != -1) {
                 $storeData = $api->ecommerce->stores->get($mailChimpStoreId);
                 $options['list_id'] = $storeData['list_id'];
-                $list = $api->lists->getLists($storeData['list_id']);
+                // The last argument asks for stats.total_contacts on this same
+                // response. It is the billable figure, and it is the only one
+                // the audience payload cannot be made to yield by arithmetic:
+                // member_count plus unsubscribe_count leaves out non-subscribed
+                // contacts entirely, which on an ecommerce account is most of
+                // the audience.
+                //
+                // Positional because that is the library's signature, and
+                // harmless against a library too old to have the parameter --
+                // PHP ignores surplus arguments to a userland function, so the
+                // field is simply absent there rather than an error.
+                $list = $api->lists->getLists(
+                    $storeData['list_id'],
+                    null,
+                    null,
+                    null,
+                    null,
+                    null,
+                    null,
+                    null,
+                    null,
+                    null,
+                    true
+                );
                 $options['list_name'] = $list['name'];
                 $options['total_list_subscribers'] = $list['stats']['member_count'];
                 $totalCustomers = $api->ecommerce->customers->getAll($mailChimpStoreId, 'total_items');


dipola (approved):
Approving. Checked the two things that could go wrong rather than reading the argument list and counting on my fingers.

**The positional slot lines up.** The library's signature is eleven parameters ending in `$includeTotalContacts`; this passes eleven ending in `true`. Verified against the signature rather than by eye:

```
signature has 11 params; 11th = includeTotalContacts
PR passes    11 args;    11th = true
ALIGNED
```

**And the older-library case is safe for the reason you give.** Surplus arguments to a *userland* function are accepted silently — only internal functions raise `ArgumentCountError` — so against a library without the parameter the field is absent, not a fatal. Confirmed on the same PHP the box runs:

```
old(1, null, null, true) -> 4 args accepted, no error
```

That is the part I would have wanted proven before this shipped ahead of the library, and you proved it both ways on a real 2.4.8 rather than in principle.

**The nine nulls are the right call**, unlovely as they look. Named arguments would need PHP 8 and this package still declares a much older floor; a `getListWithTotals()` convenience wrapper would put a second spelling of one call into a library whose whole virtue here is that it has one. Nine nulls that are obviously nine nulls beat a helper that has to be discovered.

**And the comment earns its place** by explaining the thing a future reader would otherwise delete: not *what* the eleventh argument does, which is guessable, but why the sum cannot substitute for it. 413 against 20,971 is the sentence that stops someone "simplifying" this into `member_count + unsubscribe_count` in a year's time.

One note for the record, since neither repo can state it alone: this and the library half are each inert without the other. The library can only observe what a caller asked for; this only asks. Merged apart they do nothing, which is a property worth remembering when either is considered for a release on its own.

Nothing reads the field here and the statistics payload is unchanged — so the blast radius is one query-string argument on a call that already happens, twice a day.

dipola:
Mergeable and approved — worth landing now rather than at the next release, because everything either side of it is already in place and this is the only piece that does nothing on its own.

- `ebizmarts/mailchimp-lib` **3.0.50** is tagged and on Packagist, so `observeList()` and the `include_total_contacts` parameter are out.
- The receiver went live an hour ago (edge v0.370, migration 0054): all four counts are accepted and stored, with the read-time arithmetic subtracting cleaned explicitly and both derived figures rendered as upper bounds.

So the estate is currently in the one state we agreed is the least useful: the three counts that are **not** the billable figure will start landing as installs pick up 3.0.50, and the one that is stays absent until something asks for it. Our console shows the three and deliberately serves no billable figure at all rather than a plausible sum — which is right, and also means the column stays empty until this merges.

No rush from a correctness point of view: nothing breaks while it waits, the field is simply null. It is only that the value of the other two releases is gated on one query-string argument.

Your call on the timing — merge it whenever suits the `develop-2.4` train.

PR #2366, opened 2026-09-14
Ask for total_contacts on every single-audience read, not only the cron

gonzaloebiz:
Raised in review of the library side, and the consequence is worse than a missing field.

## Why it matters

The statistics cron is not the only path reaching `lists/{id}`. **Five more do**, and all five are admin, on the **web lane — which is not sampled**, so each one emits a beacon:

```
Controller/Adminhtml/Stores/Get.php
Controller/Adminhtml/Ecommerce/Getaccountdetails.php
Model/Config/Source/Details.php
Model/Config/Source/MonkeyList.php
Helper/Data.php
```

The receiver **replaces all four audience counts on every reading, including the ones that reading does not carry.** So an admin opening the Mailchimp configuration page sent member, unsubscribe and cleaned with `total_contacts` absent — and blanked the billable figure until the next cron run, twelve hours later.

That rule was written on the assumption that `lists/{id}` was only ever reached by the twelve-hourly cron. These five are why it was not. The receiver is being hardened so a later incomplete reading cannot blank a good one, but **complete readings are the better fix on their own**: otherwise a total from one moment sits beside counts from another.

## Verified on a running install

Constructing `Model\Config\Source\MonkeyList` — the admin audience dropdown, and the shortest path to one of the five:

| | `list_member_count` | `list_total_contacts` |
|---|---|---|
| before | 3 | **null** ← blanks a good figure |
| after | 3 | **5** |

## Left alone deliberately

`Controller/Adminhtml/Lists/Get.php` fetches **every** audience rather than one. The beacon ignores collection responses — no `stats` on that shape — so asking would add payload to a response nobody reads it from.

## Sequencing

This belongs in the same release as the cron change, not after it. Shipping the cron alone would put complete readings twice a day against incomplete ones on every admin page load — which is the state the receiver rule handles worst.


dipola (changes requested):
The change itself is right and I have checked all five call sites — the arguments are in the correct position, `Lists/Get.php` is correctly left alone, and your sequencing point about shipping this in the same tag as 2365 is the most important thing on either thread.

**One thing to change before merge, and it is only the comments.**

They explain the argument in terms of what a downstream consumer does with the response, rather than in terms of this repository.

That is the problem. A reader here can verify everything about `getLists()` and `include_total_contacts` from the library source and from Mailchimp's schema. They cannot verify a consumer's storage behaviour from anything in this repository, so the comment asks them to take it on faith — and it ships to every install that pulls this package. A comment resting on an external system also goes stale silently, because nothing here can tell you when it has.

The justification is stronger without it, because it stands on this repository's own terms:

```php
// `total_contacts` is returned only when include_total_contacts is set, so a
// caller that needs it has to ask on every read. Asking only on the statistics
// cron leaves every other single-audience read returning a response without it.
```

That is checkable from the library source and Mailchimp's schema, it stays true regardless of what any consumer does, and it is the reason the argument belongs at all five sites rather than one.

Everything else is approved on sight once that is in — I do not need to see it again.

gonzaloebiz:
Rewritten — comments only, no behaviour change.

You are right on both counts, and the second is the one I should have caught myself. I wrote the same
principle into the library's `harvest()` docblock last week — describing the discriminator in terms of
the beacon rather than naming the receiver's columns, precisely so the note stays true if that storage
changes — and then did the opposite here.

All five now read:

```php
// `total_contacts` is returned only when include_total_contacts is set, so a
// caller that needs it has to ask on every read. Asking only on the statistics
// cron leaves every other single-audience read returning a response without it.
```

Nothing left in the five files mentions a consumer, its merge behaviour, or what any figure is used for
downstream.

**I checked the one already merged**, since it would have had the same problem: the comment on
`GenerateStatistics.php` justifies the argument by what the Mailchimp payload does and does not yield by
arithmetic, and by PHP's handling of surplus arguments. Both verifiable from here, so it stands as
written.

Suite unchanged, 214 tests.

The sequencing is untouched by this: same tag as 2365, or neither.

dipola (approved):
Approved.

The comments now stand on this repository's own terms at every call site: `total_contacts` is returned only when the argument is set, so a caller that needs it has to ask on every read. That is checkable from the library source and Mailchimp's schema by anyone reading this file, and it stays true regardless of what any consumer does with the response.

Nothing further from me. The one thing still worth holding to is the sequencing you raised first: this wants to reach a tag alongside #2365 rather than a release later.

PR #2367, opened 2026-09-14
Ask for total_contacts on the audience collection too

gonzaloebiz:
One argument, on the call site deliberately left out of the previous change.

## Why the earlier reasoning retired

When the five single-audience call sites were changed, `Controller/Adminhtml/Lists/Get.php` was left alone on the argument that the beacon ignores collection responses — no `stats` on that shape — so asking would add payload nobody reads it from.

That was true then. It is not now: the library is gaining the ability to observe the collection, and the collection turns out to be **the one response that can describe every audience an account has** rather than the one a store view happens to point at. It carries each audience's stats alongside the count, so it goes from ignored to richest.

Without the argument its entries arrive without the contact total — the figure that cannot be derived from the other three, since `member_count + unsubscribe_count` omits non-subscribed contacts entirely.

## Shape

Positional, like the other six, and harmless against a library too old to have the parameter: PHP accepts surplus arguments to a userland function, so the field is absent rather than fatal.

## Sequencing

This wants the same tag as the other call-site changes, for the same reason they were kept together: a release where some reads are complete and others are not is the case that goes wrong, and this is the read that covers audiences no other call site can see.


dipola (approved):
Approved.

The reasoning is right and it retires mine: I accepted that this call site could stay bare because the collection was ignored, and the collection is about to become the one response that describes every audience on an account rather than whichever one a store view points at. Without the argument its entries arrive without the figure that cannot be derived from the other three.

Positional and harmless against an older library, for the reason you give — surplus arguments to a userland function are accepted, so the field is absent rather than fatal.

Agreed on sequencing, and more strongly than on the earlier thread: a release where the collection is observed but answers without contact totals is worse than one where it is not observed at all, because the gap would look like real audiences reporting nothing rather than like a missing parameter.

PR #2368, opened 2026-09-14
version 84 for magento 2.4

gonzaloebiz:
Version bump, changelog, and the library floor.

## What it releases

- **Ask for the audience's total contacts on the call already being made** (#2365)
- **Ask for total_contacts on every single-audience read, not only the cron** (#2366)
- **Ask for total_contacts on the audience collection too** (#2367)

All seven `getLists()` call sites now ask, in one tag rather than across several. A release where some reads request the field and others do not returns responses that disagree with each other about the same audience -- one carrying the contact total, one omitting it -- and nothing in this codebase says which of the two to believe. That is checkable from the library's signature and Mailchimp's schema alone.

## The floor

`ebizmarts/mailchimp-lib` moves from `>=3.0.49` to `>=3.0.51`, and `MIN_LIB_VERSION` moves with it — a test holds the two together, since composer enforces the floor for installations installed with composer and the constant is the only thing an app/code installation is ever told.

3.0.51 is the release that reads the audience collection. Against 3.0.50 the extension asks for the collection's totals and nothing looks at the response; below that the argument is ignored outright. A floor, not a pin, as it has been since 3.0.44 — an installation already resolving to something newer is unaffected.

## Verified on the running install

Both directions of the admin warning, on 2.4.8, against real composer metadata rather than a fixture:

| library composer reports | `Hint::checkLibVersion()` |
|---|---|
| `3.0.50.0` | `'3.0.50.0'` — warns |
| `3.0.51.0` | `false` — silent |

The first row is also why the provider carries a `3.0.50.0` case: composer normalises to four segments, and that is the string the block actually receives.

Suite: **214 tests, 390 assertions, green** (`vendor/bin/phpunit -c app/code/Ebizmarts/MailChimp/phpunit.xml.dist`).

## Test fixtures that moved with the floor

- `'newer' => '3.0.50'` is now a release *below* the floor, so it changes sides and becomes the "one release behind" case.
- `'nine releases behind' => '3.0.9'` kept a count that was already wrong when written (it was forty behind the previous floor, forty-two behind this one). Renamed to say what it tests — a version with fewer digits than the floor, which is the shape the old string comparison got backwards.


dipola (approved):
Approved. The floor resolves — 3.0.51 is tagged and on Packagist — and holding `MIN_LIB_VERSION` to the composer constraint with a test is the right call, since the constant is the only thing an app/code installation is ever told.

The diff is clean and the fixture note is the good kind: `'nine releases behind' => '3.0.9'` carried a count that was wrong when it was written, and renaming it to say what it actually tests — a version with fewer digits than the floor — is worth more than correcting the number would have been.

**One line in the description to change before this is tagged**, and it is the description only, not the code.

The sequencing argument is justified by what a downstream consumer does with the response. Two problems with that, and the second is the one that matters:

1. **It is already out of date.** That behaviour changed earlier today; an incomplete reading no longer displaces a complete one.
2. **It explains this repository by reference to a system this repository cannot see**, and it ships to every installation that pulls the package.

The conclusion does not need it. Keeping the seven call sites in one tag stands on this repository's own terms: a release where some reads ask for the field and others do not returns responses that disagree with each other for the same audience, and nothing in the codebase would tell you which one to believe. That is checkable from the library source and Mailchimp's schema, and it stays true whatever any consumer does.

Same edit as on the two library descriptions — nothing else, and I do not need to see it again.

PR #2369, opened 2026-09-15
Name the surface the process is serving

gonzaloebiz:
The extension half of the area/action pair. The receiver is live and the library side is open as a separate PR; this is inert until that lands, by design.

## Why the library cannot work this out for itself

It sees `PHP_SAPI` and nothing else. But **`pub/cron.php` refuses to run under the CLI SAPI** — it exits on `php_sapi_name() === 'cli'` and then boots the framework's `Cron` application. So an installation running cron over HTTP is a web-SAPI process doing a full sync, and on the other side it is indistinguishable from a shopper's page render. Every web-lane figure is contaminated by an unknown amount of cron, and no amount of looking at our own traffic separates them.

The narrower reason. Exactly two call sites reach the campaigns family — `Controller/Campaign/Check.php`, which fires on any storefront page carrying a foreign `mc_cid`, and `Controller/Adminhtml/Orders/Campaign.php`, which fires only on an explicit click in the order view. Both emit an identical `campaigns:404` into a single last-write-wins slot, and both are web, so context cannot separate them. The storefront one is the common case by construction — it needs no merchant action at all — so suppressing it as noise is the same edit as blinding ourselves to the merchant-facing one.

## Shape

One private `applySurface()`, called from the three places that construct the API — `getApi()`, `getApiByApiKey()` and the store-refresh loop — because any of the three can be the first in a process, and the library takes the first non-empty value.

`Magento\Framework\App\State` arrives as an **optional** constructor argument named in `di.xml`, the same pairing already used for the cache. The default keeps the constructor backwards compatible; naming it in `di.xml` is what makes it arrive, since `ClassReader` records a parameter with a default as not required and `AbstractFactory::getResolvedArgument()` then hands it that default instead of instantiating its type. A test pins the two halves together, because every other test here sets the property by hand and would pass without the wiring.

## Three guards and one filter

| | why |
|---|---|
| `method_exists($this->_api, 'setSurface')` | an app/code install pairs whichever library is on disk with whichever module is on disk, so the composer constraint decides nothing there. Without it this is a fatal on every call into the API. |
| `try` around `getAreaCode()` | it throws when nothing has set an area yet, which is a normal state early in a request. Reporting no area is right; failing the API call is not. |
| `method_exists($this->_request, 'getFullActionName')` | not on `RequestInterface`, and absent from `Magento\Framework\App\Console\Request` — confirmed against 2.4 vendor source. |

**The filter is the one that matters, and it is not covered by any of the guards.** In a CLI process the object DI resolves for `RequestInterface` is `Magento\Framework\App\Request\Http`, **not** the console request — so `method_exists` is true and the guard never fires on the lane it was written for. Nothing has been routed, so all three segments are null and `getFullActionName()` returns the bare delimiters:

```
class     : Magento\Framework\App\Request\Http
method_exists getFullActionName : true
value     : '__'
```

Underscore is inside the library's charset rule and inside the receiver's, so `__` would have been **accepted, stored and counted as an action name on every cron run** — the same meaningless value on every installation that runs cron. An action has to carry at least one character that is not a separator.

## Verified on a running 2.4.8

Through the real helper, real DI, real `State` and the real request object, one run per lane:

| area | `getFullActionName()` | reported area | reported action |
|---|---|---|---|
| `crontab` | `'__'` | `crontab` | **absent** |
| `frontend` | `mailchimp_campaign_check` | `frontend` | `mailchimp_campaign_check` |
| `adminhtml` | `mailchimp_orders_campaign` | `adminhtml` | `mailchimp_orders_campaign` |

`_state` resolved to `Magento\Framework\App\State\Interceptor` in all three, which is the `di.xml` wiring doing its job rather than the default arriving.

The last two rows are the pair this field exists to separate — the storefront check and the merchant's own click, previously one indistinguishable `campaigns:404`.

**The cron lane, through the framework's own Cron application** rather than an area set by hand — this is where most of the API traffic lives, and it is the lane Reason A is about:

```
area set by Magento\Framework\App\Cron : crontab
reported                               : area='crontab'  action=NULL
```

`Cron::launch()` sets the area itself, before any SAPI question arises, so an installation running cron over HTTP reports `crontab` for the same reason — which is exactly what makes it separable from a shopper's render for the first time.

**A real dispatch through the front controller**, with the real router, the real controller and a real Mailchimp call, read out of the envelope that process would have sent:

```
routed to : 'mailchimp_campaign_check'      <- the router produced this, not the test
response  : {"valid":0}
envelope  : area=frontend  action=mailchimp_campaign_check  last_err=campaigns:404  calls=1
```

That single envelope is the whole of Reason B: the storefront's benign `campaigns:404` now arrives carrying the call site that produced it, and `Controller/Adminhtml/Orders/Campaign.php` would arrive carrying a different one.

The same URL over real HTTP against the running storefront returns `200 {"valid":0}` — the documented benign case, a visitor carrying somebody else's campaign id, answered correctly.

**Cost**, because this runs on the storefront path and that path is the subject of a separate open issue:

```
applySurface() x10000 : 6.18 ms total -> 0.0006 ms per call
```

**Second of the three call sites**, `getApiByApiKey()`: `area='frontend' action='checkout_cart_index'`.

### Not exercised

The third call site, the store-refresh loop, truncates `mailchimp_stores` before repopulating it from the API, so it was left alone on a box with data on it. It calls the same `applySurface()` as the other two and is covered by the unit tests, but it has not been run end to end.

A true authenticated admin dispatch was not made either; the `adminhtml` row above comes from a request object in the state a dispatch leaves it, not from the router.

## Tests

**225 tests, 404 assertions, green** — 11 new. Four mutations, four killed:

| mutation | result |
|---|---|
| drop the `__` filter | 3 failures |
| drop the `setSurface` guard | 1 error |
| drop the `catch` around `getAreaCode()` | 1 error |
| drop the `state` argument from `di.xml` | 1 failure |

## Not raising the library floor

Deliberately. The guard makes this safe against any library, and the feature is simply inert until one that has `setSurface()` is installed — the same way the `total_contacts` argument shipped ahead of the library that read it. Raising the floor can wait for the release that pairs them.


dipola (commented):
The `__` filter is a good catch and the reasoning behind it is right: a token that is well-formed and names nothing is worse than no token, because nothing downstream can tell it from a real one.

There is a second value in exactly that class, one line above the filter, and the filter cannot catch it.

## `(string)` is a repair, and it manufactures route names that never existed

```php
$action = (string)$this->_request->getFullActionName();
```

`getFullActionName()` is three segments concatenated, and the segments are whatever was last passed to `setRouteName()` / `setControllerName()` / `setActionName()`. Nothing constrains them to strings. Verified on a running 2.4 (framework 103.0.8, PHP 8.3), driving the real `Magento\Framework\App\Request\Http` through the ObjectManager:

```
all null (a cron)   raw_type=string  cast='__'                  passes filter: no
normal              raw_type=string  cast='checkout_index_index' passes filter: YES
array route         raw_type=string  cast='Array_index_index'    passes filter: YES
array all three     raw_type=string  cast='Array_Array_Array'    passes filter: YES
int segments        raw_type=string  cast='1_2_3'                passes filter: YES
bool segments       raw_type=string  cast='1_1_1'                passes filter: YES
```

`setRouteName(['a'])` does **not** throw — `Array to string conversion` is a warning, so execution continues — and the concatenation hands back an ordinary string.

And those setters are reachable. `module-page-cache/Controller/Block/Render.php`, same install:

```php
$origRequest = $this->getRequest()->getParam('originalRequest');
if ($origRequest !== null) {
    if ($origRequest && is_string($origRequest)) {
        $origRequest = json_decode($origRequest, true);
    }
    $this->getRequest()->setRouteName($origRequest['route']);
    $this->getRequest()->setControllerName($origRequest['controller']);
    $this->getRequest()->setActionName($origRequest['action']);
```

`class Render extends \Magento\PageCache\Controller\Block implements HttpGetActionInterface` — a frontend GET, no admin session, no form key, `json_decode` with no schema. `originalRequest={"route":["a"],...}` puts an array into `setRouteName()`.

So the reported action becomes `Array_index_index`: ASCII, well-formed, passes `/[A-Za-z0-9]/` here and the charset rule in `surfaceToken()`, and names a route that has never existed in any installation. **It was not sent by the caller — it was produced by our own cast.** That is the same defect as `__` with one difference that matters: `__` is refusable by anything that sees it, and this is not, because by then it is an unremarkable string.

It is also the one thing on this path that only this layer can fix. `surfaceToken()` already does `is_string()`, and the cast here is what defeats it.

One line, and the cast goes away with it:

```php
$action = '';
if (method_exists($this->_request, 'getFullActionName')) {
    $raw = $this->_request->getFullActionName();
    if (is_string($raw) && preg_match('/[A-Za-z0-9]/', $raw)) {
        $action = $raw;
    }
}
```

Same treatment is worth giving `(string)$this->_state->getAreaCode()` for symmetry — lower risk, since a bad area is caught by a closed set rather than a pattern, but it is the same coercion and removing both makes the rule one rule.

Fix that and this is good to go from my side.

## Two smaller things

**The `catch` is narrower than the argument for having it.** `State::getAreaCode()` throws `LocalizedException`, so the catch is correct against the class as written — but DI resolves an `Interceptor` here (your own output says so), and a plugin on `getAreaCode()` can throw anything. Your stated principle is that reporting no area is right and failing the API call is not, and a `\Throwable` catch is what actually holds that line. The library already does exactly this around its own send path, for the same reason.

**Call-site coverage is complete, and it is checkable rather than a matter of trust.** `setUserAgent()` has exactly three non-test call sites — `Helper/Data.php` 453, 652 and 1139 — and `applySurface()` is at all three. Worth stating in the docblock that the pairing is the invariant, so the next person adding a fourth API construction adds both.

## What I would not change

The optional-argument-plus-`di.xml` pairing, and the test that pins the two halves together. Every other test in the file sets the property by hand and would pass with the wiring removed — that test is the only thing standing between a green suite and every beacon silently reporting no area. Good that it exists and good that the mutation list proves it fails.

gonzaloebiz:
Fixed, and the class of defect is real — but two details of the diagnosis do not survive measurement on this platform, and one of them matters because **the patch as written does not close the hole.**

## The array vector throws before it reaches us

`setRouteName(['a'])` does not warn and continue. Same install, framework 103.0.8, PHP 8.3:

```
route = array     TypeError: Cannot access offset of type array in isset or empty
all three array   TypeError: Cannot access offset of type array in isset or empty
object            TypeError: Cannot access offset of type stdClass in isset or empty
```

`Http::setRouteName()` calls `getRouteFrontName($route)`, which indexes the route table with what it was given — `Route/Config.php:121`. So the PageCache path carrying `{"route":["a"],...}` raises inside the platform, and `Array_index_index` never composes. Not our defence; just not the vector.

## `(string)` was not manufacturing anything

Every surviving case already returns a string, because `getFullActionName()` **concatenates**:

```
all null   raw_type=string  '__'
normal     raw_type=string  'checkout_index_index'
integers   raw_type=string  '1_2_3'
booleans   raw_type=string  '1_1_1'
```

The cast was a no-op on every one. Removing it is right — a cast is a repair, and repairing untrusted input is how a value nobody sent becomes indistinguishable from one somebody did — but it was not the mechanism.

## Which is why `is_string($raw)` would not have fixed it

Your own table says `raw_type=string` in the very rows you are trying to reject. `is_string('1_2_3')` is true, and `/[A-Za-z0-9]/` passes. The patch as written leaves `1_2_3` and `1_1_1` arriving exactly as before.

**And those two are reachable.** The same PageCache controller, `{"route":1,"controller":2,"action":3}` — `setRouteName(1)` indexes the route table with `1`, which is a perfectly legal offset, so nothing throws and the concatenation hands back `1_2_3`. Booleans the same, composing `1_1_1`.

## What actually closes it

The segments, one at a time, before trusting what they composed:

```php
foreach (array('getRouteName', 'getControllerName', 'getActionName') as $getter) {
    $segment = $this->_request->$getter();
    if ($segment !== null && !is_string($segment)) {
        return '';
    }
}
```

`getRouteName()` returns `integer 1` and `boolean true` — the types survive on the getters even though the concatenation destroys them. **Null is allowed deliberately**: a part that was never routed is what makes `mailchimp__` composable, and you asked for that one to be kept.

| segments | composes | before | now |
|---|---|---|---|
| `mailchimp`, `campaign`, `check` | `mailchimp_campaign_check` | sent | sent |
| `mailchimp`, null, null | `mailchimp__` | sent | sent |
| null, null, null | `__` | dropped | dropped |
| `1`, `2`, `3` | `1_2_3` | **sent** | dropped |
| `true`, `true`, `true` | `1_1_1` | **sent** | dropped |
| `mailchimp`, `2`, `3` | `mailchimp_2_3` | **sent** | dropped |

## The other two

**`\Throwable`** — taken, and you are right about why: DI resolves an interceptor and a plugin can throw anything, so the narrow catch held the class rather than the principle.

**The call-site invariant** — stated in the docblock. `setUserAgent()` has three non-test call sites and `applySurface()` is at all three; the pairing is the thing to keep, not the count.

## Tests

**230 tests, 413 assertions, green** — 5 more. Two further mutations, both killed:

| mutation | result |
|---|---|
| drop the segment type check | 4 failures |
| narrow `\Throwable` back to `LocalizedException` | 1 error |

The three lanes re-verified live after the change: `crontab` → area only, `frontend` → `mailchimp_campaign_check`, `adminhtml` → `mailchimp_orders_campaign`.

dipola (approved):
Checking the segments rather than the composed value is the right fix, and it is better than the one I proposed. Approving.

## You are right that `is_string()` on the result was insufficient, and this was my error

`getFullActionName()` concatenates, so by the time it returns, an integer or boolean segment has already become part of an ordinary string. `is_string($raw)` cannot reject `1_2_3` or `1_1_1` — they *are* strings. I measured both of those values passing the filter and then proposed a check that does not stop them. Checking each segment, where the types still survive, is the only place the distinction exists.

## One correction going the other way: the array vector does not throw

On a 2.4 install here (framework 103.0.8, PHP 8.3.16) `setRouteName(['a'])` completes without raising anything, and `getFullActionName()` returns the string `Array_index_index`. Three arrays give `Array_Array_Array`.

The reason is narrow and worth recording, because it is the kind of thing that reads as a type error and is not one:

```php
// Magento\Framework\App\Route\Config::getRouteFrontName()
return isset($routes[$routeId]) ? $routes[$routeId]['frontName'] : $routeId;
```

`isset()` is the one array-access construct that returns `false` on an illegal offset type rather than raising. So an array `$routeId` misses the lookup, is handed straight back as `$module`, and `setModuleName()` stores it. Nothing in `setRouteName()`, `setControllerName()`, `setActionName()` or `getFullActionName()` carries a type declaration to catch it either — I checked all four signatures on that install.

So the array case is live rather than theoretical, which makes your segment check *more* load-bearing than the reasoning in your comment credits it for, not less. If you measured a `TypeError`, it is worth knowing where — a difference in Magento or PHP version that turns this from silent to loud would be useful to both of us, and it would mean the silent behaviour is confined to a version range rather than being general.

Either way it does not change the fix: your check catches arrays, integers and booleans alike, which is what it needs to do, and it does not depend on which of us is right about the exception.

## Small notes, neither blocking

**The `method_exists` sweep over all four getters is the right call.** Returning `''` if any one of them is missing means the composed value can never be built from a partially-present interface, and it keeps the guard honest on request implementations that carry some of the four.

**`$segment !== null && !is_string($segment)` allows null deliberately** — worth a word in the docblock saying so, since null is exactly the unrouted case that composes the bare delimiters, and a later reader tightening this to `is_string()` alone would reject the very case the signal filter is there to handle, one line further down.

gonzaloebiz:
Thanks for the `is_string()` correction — and on the array one, I re-measured and cannot reproduce it.

On this install, PHP **8.3.27**, `isset()` with an illegal offset **raises**:

```
isset($routes[ array ])    -> TypeError: Cannot access offset of type array in isset or empty
isset($routes[ stdClass ]) -> TypeError: Cannot access offset of type stdClass in isset or empty
```

In isolation, against an array literal, with no Magento in between. Which is why `setRouteName(['a'])` leaves through `Route/Config.php:121` before it ever reaches `setModuleName()`:

```
route = array     TypeError: Cannot access offset of type array in isset or empty
all three array   TypeError: Cannot access offset of type array in isset or empty
object            TypeError: Cannot access offset of type stdClass in isset or empty
```

`isset()` returning `false` on an illegal offset type rather than raising is **pre-8.0** behaviour. Since 8.0 it is a `TypeError`, and the message says so in as many words — *"in isset or empty"*. That error exists precisely because `isset` stopped being the exception.

You cite 8.3.16 and I am on 8.3.27, and a behaviour difference between two patch releases of the same minor would be very strange. Before either of us treats their own reading as settled, could you run just this on your box?

```php
php -r '$r = ["a" => 1]; try { var_dump(isset($r[["x"]])); } catch (\Throwable $t) { echo get_class($t), ": ", $t->getMessage(), "\n"; }'
```

**Nothing about the code changes either way**, which is why I am not raising it as blocking: an array is not a string, so the per-segment check rejects it whether it arrives as a `TypeError` or as `Array_index_index`. The difference is only whether the vector exists at all — which matters for the record, and for anywhere else someone relies on `isset` not raising.

What I do take whole is the other half: `1_2_3` and `1_1_1` were reachable, passed every rule, and a check on the composed value could not touch them.

PR #2370, opened 2026-09-15
version 85 for magento 2.4

gonzaloebiz:
Version bump, changelog, and the library floor. No behaviour changes.

## What it releases

**#2369 — the caller's surface.** The extension passes the application area and the dispatched action to the library, from the three places that construct the API.

The library sees `PHP_SAPI` and nothing else, and that answers a different question: `pub/cron.php` refuses to run under the CLI SAPI and then boots the framework's `Cron` application, so an installation running cron over HTTP is a web process doing a full sync and nothing downstream could tell it from a shopper's page render.

## The floor

`ebizmarts/mailchimp-lib` moves from `>=3.0.51` to `>=3.0.52`, and `MIN_LIB_VERSION` moves with it — a test holds the two together, since composer enforces the floor for installations installed with composer and the constant is the only thing an app/code installation is ever told.

**This is a floor for the feature, not for correctness.** The call is guarded, so against a library that has never heard of `setSurface()` it is skipped and the envelope is exactly what it was. What happens below 3.0.52 is that the installation reports no surface at all — which is the state this release exists to end, not a breakage.

A floor rather than a pin, as it has been since 3.0.44.

## Verified on the running install

The pairing, end to end, with the released library on disk and a routed request:

```
lib on disk        : 3.0.52
module             : 103.4.85
surface end to end : area='frontend' action='mailchimp_campaign_check'
```

And the admin warning, both directions, against real composer metadata rather than a fixture:

| library composer reports | `checkLibVersion()` |
|---|---|
| `3.0.51.0` | `'3.0.51.0'` — warns |
| `3.0.52.0` | `false` — silent |

The first row was the state of the box before 3.0.52 reached Packagist; the second after `composer update` moved it. Same install, same constant.

Suite: **230 tests, 413 assertions, green.**

## Test fixtures that moved with the floor

`3.0.52` was the provider's "newer" case and is now the floor itself; `3.0.50` is no longer one release behind it. Both move every time the floor moves, which is what the provider is for.


dipola (approved):
Approving. Clean release: the floor, the constant, the changelog and the fixtures all move together, and nothing else moves.

I checked the claim that a test holds the floor and the constant together rather than taking it — `testTheConstantTracksTheComposerFloor()` reads the real `composer.json` and asserts `'>=' . MIN_LIB_VERSION` against it. Reading the file rather than a fixture is what makes it a binding, and it is the right one to have: composer enforces the floor for composer installations and the constant is the only thing an app/code installation is ever told, so a drift between them would be invisible in exactly the population that has no other signal.

**The release ordering is right, and it is worth saying because it is the part that cannot be fixed afterwards.** `ebizmarts/mailchimp-lib` 3.0.52 is tagged and resolving on Packagist now, so `>=3.0.52` is satisfiable the moment this merges. Released the other way round, a fresh `composer require` of 103.4.85 would fail to resolve for however long the gap lasted.

One note rather than a request, on who actually sees the admin warning. Composer installations never will: the floor pulls 3.0.52 in as part of taking this release, so the constant and the library agree before the block ever renders. The warning exists for app/code installations, where the version constraint decides nothing — which is both the point of having it and the reason its reach is bounded. Worth keeping in mind if the warning shows up in support volume after this ships: it is a signal about install method, not about the release.

The changelog entry explaining why the segments are checked individually — integers composing `1_2_3`, booleans `1_1_1`, an unrouted request composing the delimiters — is the right level of detail for a reader deciding whether to take the release. It says what the rule refuses without requiring them to read the diff.

PR #2371, opened 2026-09-15
Give up on a batch result we cannot read, and re-send what it carried

gonzaloebiz:
A batch whose result cannot be downloaded or unpacked is retried on **every cron run** — re-fetching the status and re-downloading the archive each time. Nothing backs off and nothing gives up. What ends it is Mailchimp expiring the batch about a week later: roughly two thousand attempts at the shipped five-minute cadence, none of them counted anywhere, because that download is the one call that does not go through the library.

## Why nobody could see it

Two unrelated outcomes were the same empty array:

| situation | what `getBatchResponse()` returned | what the caller did |
|---|---|---|
| the batch has not finished | `[]` | left it pending — correct |
| finished, result unreadable | `[]` | left it pending — forever |

Indistinguishable from the outside. That is why the second was invisible, and why a naive count of attempts would have counted the first — the normal poll of a batch that is simply still running.

So `getBatchResponse()` now returns **`null`** while a batch has not finished. Waiting is the only correct thing to do there and it already ends on its own: a batch that never finishes is expired by Mailchimp, and the status call then fails in a way that was handled all along.

**That is not additive for every caller, and the first version of this PR said it was.** `null` does fall through both checks — and then reaches `foreach ($files as $file)` in the two admin download controllers, which guard on `$files === false` only. Corrected below.

## The cap counts only what is already wrong

Five readings of a **finished** batch. The work completed on the other side, so the condition is known to be wrong from the first attempt — this is not a retry budget for an operation that might still succeed on its own.

Five at the shipped `*/5 * * * *` is about twenty-five minutes: long enough for a transient network or storage fault, short enough not to spend a week on a result that will never parse. A test pins the number **and** the cadence, since the first is meaningless without the second.

## Giving up re-sends rather than deletes

The operations already ran on Mailchimp, so what is lost is the record of which of them failed — and asking again cannot recover that once the result is unreadable. What the batch carried goes again, found by the batch id the sync rows still hold.

**Marked, not deleted.** And marked with three fields rather than one, which is worth spelling out because this repository contains two different versions of "mark as modified" that disagree:

```php
markAllAsModified()       // mailchimp_sync_modified, mailchimp_sent = NEEDTORESYNC, batch_id = null
markAllAsModifiedByIds()  // mailchimp_sync_modified
```

`Model/Api/Product.php:328` selects on `mailchimp_sent = NEEDTORESYNC`, so the one-field version would have left products behind while re-sending everything else. The new by-batch-id method follows the three-field recipe.

## The cause is logged

It was discarded in favour of a fixed string. The three lines that landed named a missing directory left over from the cleanup — pointing whoever read them at the wrong thing entirely:

```
Something went wrong retrieving result for batch [X]
Deleting temporary files, will retry the next run don't worry
FilesystemIterator::__construct(.../var/Mailchimp/X): Failed to open directory
```

Now the first line carries `$e->getMessage()`.

## Verified on a running 2.4.8, against the real API

A real batch, submitted with an operation certain to fail, then made unreadable:

```
A) not finished yet   -> status=pending  attempts=0     <- the normal poll is not counted
   batch finishes, errored_operations=1
B) run 1              -> status=pending  attempts=1
   run 2              -> status=pending  attempts=2
   run 3              -> status=pending  attempts=3
   run 4              -> status=pending  attempts=4
   run 5              -> status=error    attempts=5

entities the batch carried, afterwards:
   id=777701  modified=1  sent=2  batch_id=NULL
   id=777702  modified=1  sent=2  batch_id=NULL
```

`sent=2` is `NEEDTORESYNC`. Test rows removed afterwards; the install is back to its prior state.

## Schema

One column, `response_attempts`, a `smallint` defaulting to `0` on `mailchimp_sync_batches`. It cannot be derived from what is already there: `modified_date` measures time rather than attempts, and the failing path deliberately does not write to the row, so using it would have destroyed the submission time it currently preserves.

## The two download buttons

`Controller/Adminhtml/Batch/GetResponse.php` and `Controller/Adminhtml/Errors/Getresponse.php` both read `getBatchResponse()`, guard on `false` only, and then iterate. Both wrap that in a `do/while` that re-runs while `$fileContent` is empty, and both return a `TYPE_JSON` **download**.

Each now has its own branch, because the states are not equivalent — "deleted from Mailchimp servers" is final, "has not finished yet" resolves itself and is the common case for someone clicking that button. It could not be said at all until `getBatchResponse()` separated the two.

Executed through the real controller with a doubled `Result`:

```
empty array -> 0 warnings   body []
false       -> 0 warnings   body "Response was deleted from MailChimp servers"
null        -> 0 warnings   body "The batch has not finished on Mailchimp yet, try again…"

without the null branch  -> 5 warnings, body []
                            "foreach() argument must be of type array|object, null given"
without the break        -> TypeError: count('The batch has n…')
```

Five is `MAX_RETRIES`. And the second line is the one worth reading twice: both branches assign a **string** to `$fileContent`, and the loop's condition is `while (!count($fileContent) …)`. A branch that assigns without leaving calls `count()` on a string — a TypeError on PHP 8, so the button returns a 500 rather than a corrupt file. The pre-existing `break` was never stylistic.

## Tests

**242 tests, 456 assertions, green.**

| mutation | result |
|---|---|
| raise the cap to 5000 | 1 failure |
| give up by deleting instead of marking | 1 failure |
| do not persist an attempt that does not give up | 1 failure |
| drop a controller's `null` branch | 2 failures each |
| drop only the `break;` from a branch | 1 failure |

The last one is the reason this section is worth reading. The controller tests originally asserted that both guards appear before the loop — which failed when the whole branch went, and stayed **green** when only the `break;` did. That mutation reintroduces the defect, so the test passed against the bug it existed to prevent.

**What the green number does not cover.** Both controllers are backend actions needing a full `Backend\App\Action\Context` to construct, so the suite reads their source rather than executing them: it pins the guards. The behaviour above was measured with a probe on a running install, not by the suite.


gonzaloebiz:
Pushed one commit that should have been in the first: `testTheCapAndTheCadenceThatGivesItMeaning`. It was written with the rest and dropped by a bad rebase on my end, so the branch as originally pushed had 235 tests rather than the 236 the description claims — and, more to the point, the mutation the description reports as killed (raising the cap to 5000) would have survived it.

No review had started, so nothing was reviewed against the wrong tree. The count in the description is now accurate.

dipola (commented):
The defect is real and the diagnosis is the good part: two unrelated outcomes collapsing into one empty array is why nobody could see it, and splitting "not finished" from "unreadable" is the right fix rather than a counter bolted onto the symptom.

**But the backward-compatibility claim does not hold for two of the three callers, and neither is in this diff.**

The description says callers doing `$files === false` "behave exactly as before, because `null` falls through". `null` does fall through the check — and then reaches a `foreach`:

`Controller/Adminhtml/Batch/GetResponse.php`, and identically `Controller/Adminhtml/Errors/Getresponse.php`:

```php
do {
    $counter++;
    $files = $this->_result->getBatchResponse($batchId, $batches->getStoreId());
    if ($files===false) {
        $fileContent = "Response was deleted from MailChimp servers";
        break;
    }
    foreach ($files as &$file) {          // <- null arrives here
        …
        $fileContent[] = $content;
    }
    …
} while (!count($fileContent) && $counter<self::MAX_RETRIES);
```

Before this PR an unfinished batch returned `[]`, the `foreach` iterated zero times and nothing was emitted. After it, `null` reaches the same line and PHP 8 raises

```
Warning: foreach() argument must be of type array|object, null given
```

— and because `$fileContent` stays empty, the `while` runs it again. So it is **`MAX_RETRIES` warnings per click**, not one.

The loop spinning on an unfinished batch is pre-existing and not yours. The warnings are new, and they matter more than they look: both controllers return a `ResultFactory::TYPE_JSON` download. On any install with `display_errors` on, warning text in the output stream corrupts the JSON the merchant downloads — a broken file rather than an empty one, from a button that used to work.

One line in each:

```php
if (!is_array($files)) {
```

which keeps the `false` case reaching the same message and stops `null` reaching the loop. If the two states deserve different text — "deleted from Mailchimp" versus "not finished yet" — that is worth having too, since the second is now knowable for the first time and is the more common thing a merchant clicking that button will hit.

Worth a test on each, for the same reason the rest of this PR is tested: the change that reintroduces this is a one-character edit to a comparison, and nothing currently fails if someone makes it.

Everything else reads correctly to me — the five-reading cap pinned together with the cadence, re-sending rather than deleting, and using the three-field `markAllAsModified` recipe rather than the one-field variant that would have stranded products behind `mailchimp_sent = NEEDTORESYNC`.

dipola (changes requested):
Escalating my earlier comment to a request, because it is a regression this PR introduces in two files the PR does not touch, which is the kind that gets merged past.

Unchanged from what I wrote above: `getBatchResponse()` returning `null` for an unfinished batch reaches `foreach ($files as &$file)` in both `Controller/Adminhtml/Batch/GetResponse.php` and `Controller/Adminhtml/Errors/Getresponse.php`, because their guard is `$files === false`. PHP 8 warns, `$fileContent` stays empty, the `do/while` runs it again — `MAX_RETRIES` warnings per click rather than one — and both controllers return a JSON **download**, so with `display_errors` on the merchant gets a corrupt file from a button that used to work.

One line in each: `if (!is_array($files))`.

Everything else in the PR I would take as it stands. The diagnosis is the valuable half and the fix is right: splitting "not finished" from "unreadable" rather than counting the symptom, the cap pinned together with the cadence, re-sending rather than deleting, and the three-field `markAllAsModified` recipe instead of the variant that would have stranded products behind `mailchimp_sent = NEEDTORESYNC`.

gonzaloebiz:
Fixed in both, and you were right that it is the kind that gets merged past — the description's backward-compatibility claim was about the `=== false` check and I never followed the value past it.

Verified before changing anything: both controllers guard on `$files === false` and then iterate, both wrap it in a `do/while` that re-runs while `$fileContent` is empty, and both return a `TYPE_JSON` download with a `Content-disposition` attachment header. So it is `MAX_RETRIES` warnings into the body of a file the merchant saves.

## Two branches rather than `!is_array()`

Taking the second half of your note. The states are not equivalent: **"deleted from Mailchimp servers" is final, "has not finished yet" resolves itself** — and it is the common one for someone clicking that button, which until `getBatchResponse()` separated them could not be said at all.

```php
if ($files === false) {
    $fileContent = "Response was deleted from MailChimp servers";
    break;
}
if ($files === null) {
    $fileContent = "The batch has not finished on Mailchimp yet, try again in a few minutes";
    break;
}
```

Breaking on `null` also stops the loop re-running something that cannot change inside one request. The spinning was pre-existing and not mine, but the branch that fixes the warning is the same line that would have to skip it, and leaving it spinning would be choosing to.

## The tests read the source

Both are backend actions needing a full `Backend\App\Action\Context` to construct, so executing them buys a lot of scaffolding for one branch. They assert that neither `$files === false` nor `$files === null` has left the code before `foreach ($files as`, and that the two messages are still distinct.

One mutation per controller — removing its `null` branch — fails **two** assertions each.

They also caught something while I wrote them: the two controllers differ by one character, `foreach ($files as &$file)` in one and `foreach ($files as $file)` in the other. My first version hardcoded the by-reference form and passed on one file while silently not testing the other.

**240 tests, 448 assertions, green.**

gonzaloebiz:
Correction to my last comment: **the tests I reported as covering this did not cover it.**

They asserted both guards appear before the loop. Removing the whole `null` branch failed them — that is the mutation I ran and reported. But removing only its `break;` left the suite **green**, and that mutation reintroduces the defect.

So I executed the controller instead of reading it.

## What it actually does

Real controller, real `do/while`, a doubled `Result`, warnings counted rather than inferred:

```
empty array -> 0 warnings   body []
false       -> 0 warnings   body "Response was deleted from MailChimp servers"
null        -> 0 warnings   body "The batch has not finished on Mailchimp yet, try again…"
```

And the two mutations:

```
without the null branch  -> 5 warnings, body []
                            "foreach() argument must be of type array|object, null given"
without the break        -> TypeError: count('The batch has n…')
```

**Five is `MAX_RETRIES`**, exactly as you said — I had that from your description and now it is measured.

## The second one is worse than the defect being fixed, and I had not seen it

Both branches assign a **string** to `$fileContent`, and the loop's condition is `while (!count($fileContent) && …)`. A branch that assigns without leaving therefore calls `count()` on a string, which on PHP 8 is a TypeError, not a warning. The button returns a **500** rather than a corrupt file.

That is why the pre-existing `false` branch has always had its `break` — it is load-bearing for the shape of the loop, not a stylistic early exit. The new branch inherits the same requirement, and nothing said so until it was removed and run.

A test now asserts each branch contains a `break;`, and the fine mutation fails it.

**242 tests, 456 assertions, green.**

The source-reading approach stays for what it is good for — these are backend actions needing a full context to construct, so the executed evidence above comes from a probe on a running install rather than from the suite. Stated so nobody reads the suite as proving the behaviour: it pins the guards, and the numbers above are what pins what they prevent.

dipola (approved):
Both controllers, both with the `break`, and the two states given different text — "not finished, try again" is the one a merchant clicking that button will actually hit, and it was unsayable before this PR made the states distinguishable.

**The correction is the substantive part of this round, and it is worth naming precisely.** Your tests asserted both guards appear before the loop. Removing the whole `null` branch failed them; removing only its `break;` left the suite green — and that mutation reintroduces the defect. A test that passes against the defect is worth less than no test, because it converts an open question into a settled one. Finding it by mutating finer rather than by re-reading is the only way that surfaces.

**And the thing underneath is worse than what I reported.** I said a warning in the output stream corrupts a JSON download. You found that a branch assigning without leaving calls `count()` on a **string**, which on PHP 8 is a TypeError — a 500, not a corrupt file:

```
without the null branch  -> 5 warnings, body []
without the break        -> TypeError: count('The batch has n…')
```

So the pre-existing `false` branch's `break` was never stylistic. It is load-bearing for the shape of the loop, and nothing in the file said so until it was removed and run. That is now pinned by a test asserting each branch contains one, which is the guard that would have caught this round's near-miss.

Your five warnings is `MAX_RETRIES` measured rather than reasoned, which is the half of my review I could only assert.

One note for the record rather than a request: the executed evidence comes from a probe on a running install, not from the suite, and you said so rather than letting the green count imply otherwise. Worth keeping that sentence in the PR — these are backend actions needing a full context to construct, and the next person to read `242 tests, green` should know what that number does and does not cover.

PR #2372, opened 2026-09-15
Carry the configuration snapshot on the extension beacon

gonzaloebiz:
Step one of retiring the support-log lane: **move first, retire second.** Nothing is removed here.

## Why it has to move before anything is removed

The lane that carries this today also carries full request and response bodies — which on a customer sync is the shopper's email, name and address — and it writes them into the merchant's own database until they hit the `TEXT` ceiling and are silently truncated at 64 KB. Retiring it is agreed. But the configuration snapshot is the one thing in it that nothing else replaces, and for *"this merchant says X is not syncing"* it answers more than anything else we hold.

So it lands on the beacon first, is verified arriving, and only then does anything get switched off.

## Shape

A fourth block on the ping body, beside `versionBlock()` and the liveness signals. **Flat scalars, one key per setting** — the receiving whitelist takes a type and a cap per key and has no path into a nested object, so a nested `config` object would arrive, be accepted, and be dropped.

**The contract is 24 keys**: 22 always, plus `cfg_popup_url` only when the popup is on and `cfg_delete_action` only when two-way sync is on. So an install emits 22, 23 or 24.

Measured on a running 2.4.8: **23 keys, 605 bytes**, against a 32 KB envelope cap — that is one box with the popup off, a sample rather than the spec. No value is anything but a scalar or null.

## No merchant-authored text

Two of the settings are HTML the merchant writes for the subscription success page. They are reported as **whether they were customised**, never as their content.

Not only because they would exceed the per-string cap. The receiver **truncates** an over-length string rather than refusing it, so real HTML would be stored as a 512-byte fragment ending mid-tag — something that looks like content, is not, and reads as real a year later.

## ⚠ Two settings change scope, and the comparison should expect it

The lane this replaces reads two of these at default scope while every other line in the same block passes the store id:

```php
$config['general']['delete_action'] = … getConfigValue(Data::XML_PATH_WEBHOOK_DELETE);   // no $storeId
$config['general']['show_groups']   = … getConfigValue(Data::XML_INTEREST_IN_SUCCESS);   // no $storeId
```

A multi-store-view install has been reporting the default's value on every view. Fixed here rather than carried across — which means that while both lanes run, they will **legitimately disagree** on `cfg_delete_action` and `cfg_show_groups`, on those installs only. Anything else diverging is a real defect in the port.

## What is deliberately not here

**The field map.** `Helper\Data::getMapFields()` memoizes into `$this->_mapFields` with no store id in the condition, so the first call wins for the life of the object. A per-store-view key fed by a store-blind function would report the first view's map on every row — and the parity check would *pass*, because both lanes would be wrong in the same way.

That is a sync defect rather than a telemetry one: `Cron/Ecommerce.php:128` loops every store view with one helper instance, and the subscriber sync reaches the same function through `getMergeVarsBySubscriber()`. It is being fixed separately, and the map lands after it, when the key would mean something.

## Inert until the receiving side knows the keys

The projector is a generative whitelist — it builds its output from a field table rather than copying the body — so keys it does not know are discarded while the beacon still answers 202. **These 23 keys do nothing until they are whitelisted.** That is why this PR claims no end-to-end verification: the emitter is proven, the round trip is not, and it cannot be from this side.

## Tests

**238 tests, 477 assertions, green** — 8 new, covering the flat-scalar contract, both scope fixes with a control from the same block, the HTML fields as presence, the two toggle-dependent keys, empty-versus-absent, and the type mapping.


dipola (approved):
Approving. Flat scalars, the two HTML fields as presence, the two default-scope reads fixed, no field map until 2373 lands.

**One correction to the description rather than to the code.** It says "23 flat scalar keys", measured at 605 bytes on a running install. The code emits **22 unconditionally**, plus `cfg_popup_url` only when `cfg_popup_form` is on and `cfg_delete_action` only when `cfg_two_way_sync` is on:

```php
if ($snapshot['cfg_popup_form']) {
    $snapshot['cfg_popup_url'] = …
}
if ($snapshot['cfg_two_way_sync']) {
    $snapshot['cfg_delete_action'] = …
}
```

So the contract is **24** and any given install emits 22, 23 or 24. The 23 is a true measurement of one box with one toggle off — a sample, not a spec, and the distinction matters for anyone sizing against that sentence rather than reading `forStore()`. Worth a word in the description so the next reader does not take 23 as the contract.

The conditional emission itself is the right call and I would not change it: a key present with an empty value would claim the setting was read and found blank, which is not what happened.

`assertArrayNotHasKey` on both in the gate-off test is what makes that a contract rather than an implementation detail. Good that it is asserted in both directions.

The consuming end is ready for these, so the round trip becomes verifiable as soon as this is released — nothing further needed from your side to prove it.

gonzaloebiz:
Corrected in the description — counted from `forStore()` rather than from either of our recollections: **22 unconditional, 2 conditional, contract of 24**, and an install emits 22, 23 or 24. The 605 bytes stays as what it is, one box with the popup off.

You are right that it matters: anyone sizing the whitelist against "23" would be a key short, and the key they would be short of is one of the two that only appear when a feature is on — so it would look fine on most installs and not on the ones that use it.

PR #2373, opened 2026-09-15
The merge-field map is per store view, not per process

gonzaloebiz:
`Helper\Data::getMapFields()` memoized into one property with no store id in the condition:

```php
private $_mapFields = null;

public function getMapFields($storeId = null, $options = true)
{
    if (!$this->_mapFields) {        // no $storeId, no $options
```

First call wins for the life of the object, whatever arrives afterwards.

## Why that reaches a merchant's audience

`Cron/Ecommerce.php:128` loops **every store view with one helper instance**, and the subscriber sync inside that loop reaches this through `getMergeVarsBySubscriber()` → `getMergeVars()` → `getMapFields($storeId)`.

So on a multi-store-view install with different maps per view, **every view after the first synchronises with the first view's field map.** Wrong merge fields written into the merchant's audience, silently — nothing errors, nothing is logged, and no counter moves.

## Somebody had already met this

```php
Helper/Data.php:657   public function resetMapFields() { $this->_mapFields = null; }
```

One caller in the whole codebase: `Model/Api/Customer.php:111`, clearing the memo at the top of the customer sync's per-store batch. The subscriber lane never got it.

That is a workaround for this bug, applied to one of the two consumers. It goes with the bug — left in place, its single caller would keep implying the memo is unsafe to trust.

## `$options` is in the key too

The less obvious half. It changes what each entry carries:

```php
'options' => $options ? $customerAtt[$customerFieldId]['options'] : false
```

Two callers ask with it (`getMergeVars`, `Cron/Webhook`) and one without (`Cron/GenerateStatistics`). `bin/magento cron:run` executes every due job of a group **in one process**, so the variants meet.

## Two smaller things in the same three lines

**The property was `null`, not `[]`.** Callers iterate the result, and an install with nothing mapped handed them a null.

**The condition was a truthiness test.** An empty map is falsy, so it was never remembered: `getBindableAttributes()` plus the unserialise ran again on every call, for every subscriber in the batch — the memo failing to fire in exactly the case it is cheapest to serve.

**A map that fails to parse is deliberately not memoized.** That is not an empty map, and remembering it would make one bad read permanent for the process.

## Tests

**235 tests, 421 assertions, green** — 6 new. Three mutations, three killed:

| mutation | result |
|---|---|
| drop the store id from the key | 1 failure — the second store view gets the first one's map |
| drop `$options` from the key | 1 failure |
| truthiness instead of `array_key_exists` | 1 failure — the empty map is re-read every call |

## Not verified on a live multi-store-view install

Stated plainly. My install has one store view and no map configured, so the memo never engages there and the defect cannot be reproduced on it. What is above is source plus unit tests, and the strongest independent evidence is the `resetMapFields()` workaround: somebody hit this and patched one lane.

## Why it is separate

It unblocks the field-map keys on #2372, but it is a sync defect rather than a telemetry one — merchant-affecting, different urgency, and it has to be revertable without taking the configuration snapshot with it.


dipola (approved):
Approving. Keyed by store view **and** `$options`, `array_key_exists` rather than truthiness, and `resetMapFields()` removed with its caller rather than left as a workaround outliving its bug.

The detail I would not have thought to ask for is the best one here:

```php
} catch (\Exception $e) {
    $this->log($e->getMessage());
    // Not memoized: a map that failed to parse is not an empty map
    return $mapFields;
}
```

A parse failure caching as an empty map would make one bad read permanent for the process, and it would look exactly like an install with nothing mapped. Distinguishing those two is the whole reason `array_key_exists` is in there, and caching the failure would have quietly undone it.

## One layer down, same pattern — worth checking before this closes

`getMapFields()` calls `getBindableAttributes()`, which is `getCustomerAtts() + getAddressAtt()`. Both of those memoize the way this one used to:

```php
private function getCustomerAtts()
{
    $ret = [];
    if (!$this->customerAtt) {      // truthiness, no key
```

The attribute *definitions* are global, so most of that structure is genuinely store-independent and a shared memo is correct. **The part I am not sure about is `options`.**

`getCustomerAtts()` builds each entry's `options` from `$item->getSource()->getAllOptions()`, and attribute option **labels** in Magento are store-scoped — that is what per-store option labels are for. `Cron/Ecommerce` calls `setCurrentStore($storeId)` inside the same loop this fix is about, which suggests something in there does depend on the current store.

If that is right, then after this PR the map's *shape* is correctly per store view while the option *labels* inside it still come from whichever store view was current when the first call landed — the same defect, one level down, on a narrower surface. It would only show on a multi-store-view install with translated option labels on a mapped select attribute.

**I have not verified it** and I am not asserting it — `getAllOptions()` scoping is the part I would want checked rather than reasoned about, and you have the install. Flagging it because this PR is the natural moment: the same two lines would take a key, and finding it later means a second round through the same function.

Not blocking. The fix in front of us is correct on its own terms and the case above is strictly narrower than the one it closes.

gonzaloebiz:
Checked it rather than reasoned about it, and **you are right — with one qualification that narrows it a lot.**

## The scoping is real

`Magento\Eav\Model\Entity\Attribute\Source\Table::getAllOptions()`:

```php
$storeId = $this->getAttribute()->getStoreId();
if ($storeId === null) {
    $storeId = $this->storeManager->getStore()->getId();   // the CURRENT store
}
…
->setStoreFilter($storeId)
```

So labels are store-scoped, and when the attribute carries no store id the source falls back to whichever store view is current at that moment. Measured on the install — every source-backed customer attribute returns null:

```
attributes with a source : 4
with getStoreId() null   : 4        <- all of them fall back to the current store
  website_id   source=Customer\Attribute\Source\Website
  store_id     source=Customer\Attribute\Source\Store
  group_id     source=Customer\Attribute\Source\Group
  gender       source=Eav\Entity\Attribute\Source\Table
```

And `Cron/Ecommerce.php:134` calls `setCurrentStore($storeId)` inside the same loop, so the current store genuinely differs between calls. With `getCustomerAtts()` memoized on truthiness and no key, the first store view's labels are what every later view gets.

## The qualification

**Only one of the four reads store-scoped option labels.** `Website`, `Store` and `Group` are custom sources that build their lists from websites, stores and customer groups — they do not go through `setStoreFilter` on an option collection at all. The one that does is `gender`, plus any select attribute a merchant adds.

So the blast radius is: a multi-store-view install, with translated option labels, on a mapped select attribute. Narrower than the map bug this PR closes, exactly as you said, and now with a number on it — one stock attribute.

## What I would do

Fold it in here rather than open a second round through the same function, unless you would rather keep this PR to the one defect. The change is not symmetrical with the one already made, though, and that is worth knowing before choosing: the map's key is an explicit `$storeId` argument, while `getCustomerAtts()` has no store parameter at all — its scope arrives implicitly through the store manager. So keying it means either threading a store id down from `getMapFields()`, or keying on `storeManager->getStore()->getId()`, which is a memo whose key is a global.

I prefer threading the id: the implicit version would be correct and would still read as if it could not be trusted, which is the thing we just removed with `resetMapFields()`.

Say which and I will push it here or file it.

dipola:
**Fold it in here, and thread the id.** The call graph makes the second part easier than your note assumes.

## Why fold rather than file

It is the same bug, not a neighbouring one. Ship half of it and the next reader sees `getMapFields()` correctly keyed by store view and reasonably concludes the data it composes is too. That is a worse state than either the bug or the fix alone, because it is the one that stops people looking.

Your own narrowing is what makes this cheap: one stock attribute plus merchant-added selects, on multi-store-view installs with translated labels. Small enough to carry, too small to be worth its own round through the same function.

## Threading costs nothing here — the whole chain is private and single-caller

```
getMapFields($storeId)          <- has the id already
  └─ getBindableAttributes()    <- 1 caller: Helper/Data.php:664
       ├─ getCustomerAtts()     <- 1 caller: :579
       └─ getAddressAtt()       <- 1 caller: :580
```

Three private signatures, one call site each, and **no consumer outside the helper sees a change**. Worth noting explicitly because it looks like it might: `Block/Adminhtml/System/Config/Form/Field/VarsMap.php` calls `_getAddressAtt()` and `_getBindableAttributes()` — different methods on a different class, underscore-prefixed, not these. So there is no second consumer to keep compatible and no default argument to add for one.

That removes the tradeoff you were weighing. Threading is not the more expensive option here, it is the cheaper one.

And your reason for preferring it stands on its own: a memo keyed on `storeManager->getStore()->getId()` is keyed on mutable global state, so the key can change without any caller having asked for a different thing. It would be correct and would read as though it could not be trusted — which is what removing `resetMapFields()` just fixed.

## One more in the same breath

`getAddressAtt()` memoizes identically — `if (!$this->addressAtt)` at :625, truthiness, no key — and your measurement covered the customer side only. If any address attribute is source-backed it has the same exposure; if none is, the shared memo is provably fine and worth a line saying so. Either way it is one check while you are in the function, and it is on the same call path.

PR #2374, opened 2026-09-15
The attribute options are per store view too

gonzaloebiz:
Follow-up to #2373, raised in its review. Same call path, one layer down.

## What it is

`Magento\Eav\Model\Entity\Attribute\Source\Table::getAllOptions()`:

```php
$storeId = $this->getAttribute()->getStoreId();
if ($storeId === null) {
    $storeId = $this->storeManager->getStore()->getId();   // the current store
}
… ->setStoreFilter($storeId)
```

Option labels are store-scoped, and an attribute carrying no store id takes that fallback. Measured on a running install — **every** source-backed customer attribute returns null:

```
attributes with a source : 4
with getStoreId() null   : 4
  website_id  store_id  group_id   (sources that build from websites/stores/groups)
  gender                            (Eav\Entity\Attribute\Source\Table)
```

`Cron/Ecommerce.php:134` sets a different current store on every pass of the loop #2373 was about. So after that fix the map's **shape** was correctly per store view while the option **labels** inside it stayed frozen at whichever view resolved first.

**Narrower than the bug it follows.** Only attributes with a source are exposed, and of the four only `gender` reads store-scoped option labels — the other three build their lists from websites, stores and customer groups. A merchant's own select attributes are the rest of the exposure.

## Keying alone would not have fixed it

The store id is also **set on the attribute** before its source is asked, so the labels come from the store view that was requested rather than from whichever one happened to be current. Without that the memo would be keyed by one thing and filled from another — worse than not keying it, because it looks deliberate.

```
gender before : getStoreId()=NULL
gender after  : getStoreId()=1     <- no longer takes the fallback
```

## Threaded, not read from the store manager

A memo keyed on `storeManager->getStore()->getId()` is keyed on mutable global state: the key can change without any caller having asked for a different thing. It would be correct and would still read as though it could not be trusted — which is what removing `resetMapFields()` fixed in #2373.

The whole chain is private with one caller each, so **nothing outside the helper sees a signature change**:

```
getMapFields($storeId)        has the id already
  └─ getBindableAttributes()  1 caller, Helper/Data.php
       ├─ getCustomerAtts()   1 caller
       └─ getAddressAtt()     1 caller
```

`Block/.../VarsMap.php` has its own `_getBindableAttributes()` and `_getAddressAtt()` — different class, underscore-prefixed, not these.

## `getAddressAtt()` deliberately does not take a store

Checked rather than assumed: it composes a fixed list of fourteen address element names and gives every one of them `'options' => []`. There is nothing in it a store view could change, so its shared memo is correct. The reason is now written where the next reader will ask.

## Tests

**244 tests, 486 assertions, green.**

One mutation, and it is worth reporting how it went. The first version of the new test asserted that `getCustomerAtts()` *takes* a `$storeId` parameter — and removing the argument from the call site left the suite **green**, because a parameter existing says nothing about anything passing one.

It now seeds two different attribute tables under two memo keys and reads which one the composed map came from. The same mutation fails on the value produced:

| mutation | before | after |
|---|---|---|
| drop `$storeId` from the `getBindableAttributes()` call | passed | 1 failure |

## Not verified on a live multi-store-view install

Same limit as #2373: one store view here, so the frozen-label case cannot be reproduced on it. What is above is Magento's source, a measurement of the null store ids, and unit tests.


dipola (approved):
Approving. Threaded rather than read from the store manager, `customerAtt` keyed and tested with `array_key_exists` like #2373, and `getAddressAtt()` answered rather than left ambiguous.

**The part I did not anticipate is the one that mattered**, and it is worth having in the record rather than only in the diff:

> Keying alone would not have fixed it. The store id is also **set on the attribute** before its source is asked … Without that the memo would be keyed by one thing and filled from another — worse than not keying it, because it looks deliberate.

That is right, and my review only asked for the key. A memo keyed per store view and populated from whatever store happened to be current would have passed every test anyone would think to write against the key, while still returning the first view's labels. Verifying the fallback actually stops being taken —

```
gender before : getStoreId()=NULL
gender after  : getStoreId()=1
```

— is what makes the difference between the two observable at all.

`if ($storeId !== null)` around the `setStoreId` is the right guard: a caller that names no store keeps the old fallback rather than being silently pinned to a store it did not ask for.

**On `getAddressAtt()`** — a fixed list of address element names, every entry with an empty `options`, so nothing a store view can change. That is exactly the answer worth writing down, and the comment saying *why* one of the two takes a store and the other does not is the part that stops this being re-litigated. A shared memo that is provably correct reads like an oversight unless something says otherwise.

One note rather than a request: `setStoreId()` mutates the attribute model, and it is safe here because `$this->_attCollection->create()` builds a fresh collection on every call that reaches it — the memo returns before that line on a hit. Worth knowing as a constraint if the collection is ever hoisted or cached for performance, because that change would silently reintroduce cross-store leakage through a different door.

gonzaloebiz:
Took the note into the code rather than leaving it in the thread — by your own argument about `getAddressAtt()`: a constraint that is obvious while writing it is invisible to whoever makes the change that breaks it, and this one breaks silently.

```php
// Mutates the attribute model, which is safe only because the collection
// above is built fresh on every call that reaches this line -- a memo hit
// returns before it. Hoisting or caching that collection for performance
// would reintroduce cross-store leakage through a different door, and
// nothing would fail when it did.
```

Comment only, suite unchanged at 244/486. Checked that `dismiss_stale_reviews` is off before pushing to an approved PR, so your approval stands rather than being silently dropped.

PR #2375, opened 2026-09-15
Carry the merge-field map and the interest groups

gonzaloebiz:
The half of #2372 that waited on #2373. The map is per store view now, so these keys mean something — before that fix, a per-store-view key fed by a store-blind function would have reported the first view's map on every row, and the parity check would have *passed* because both lanes were wrong the same way.

## Pairs, not counts

A count cannot answer the question the snapshot exists for. *"Twelve fields mapped"* does not say whether `firstname → FNAME` is one of them, and *"first name is not syncing"* is the ticket this is meant to settle.

Both halves of each pair are identifiers — a Magento attribute code and a Mailchimp merge tag. Schema rather than merchant content.

```
cfg_field_map    firstname:FNAME,lastname:LNAME
cfg_field_map_n  2
cfg_interest     abc123,def456
cfg_interest_n   2
```

## Bounded by whole entries, never by cutting the string

The receiving side **truncates** an over-length value rather than refusing it. A list cut mid-pair would arrive looking like a complete map that happens to end on `firstna` — so entries are dropped whole, one at a time, against the serialised length rather than a fixed entry count.

**The offered count travels beside the list and is never the trimmed one**, so a shortened map is visibly shortened rather than quietly short.

## Interest groups come from configuration

`XML_INTEREST` already stores them as a comma-separated list of ids. Read directly, **never** through `Helper::getInterest()`, which resolves them against Mailchimp — a call this must never make.

## Contract

**28 keys**: 24 always, 4 conditional. Counted from `forStore()` rather than from recollection, after the 23-vs-24 correction on #2372.

| conditional | present when |
|---|---|
| `cfg_popup_url` | `cfg_popup_form` is true |
| `cfg_delete_action` | `cfg_two_way_sync` is true |
| `cfg_field_map` | `cfg_field_map_n` is not zero |
| `cfg_interest` | `cfg_interest_n` is not zero |

Measured on a running install: 25 keys, 644 bytes, against the 32 KB cap. That box has nothing mapped, so both counts are zero and both lists are absent — a sample, not the spec.

## ⚠ A question for the receiving side rather than a claim

The merge rule written for the first two conditionals is *"a gate reported **false** clears its dependent; a gate not reported at all leaves the snapshot untouched."* That works because those gates are booleans.

**These two gate on a number.** `cfg_field_map_n: 0` is the same statement as `cfg_popup_form: false` — the merchant unmapped everything and the stored map should be cleared — but it is not literally `false`, and a rule keyed on falsity would need `0` to count.

If it does not, a map from before the merchant cleared it sits beside a count of zero forever, which is exactly the misreading the pairing exists to prevent. Worth confirming rather than assuming; happy to emit an explicit boolean gate instead if that is cleaner at your end.

## Tests

**260 tests, 576 assertions, green** — 4 new. Two mutations, two killed:

| mutation | result |
|---|---|
| trim by characters instead of whole entries | 1 failure — a pair arrives cut in half |
| report the trimmed count instead of the offered one | 1 failure |


gonzaloebiz:
Re-tested properly, and the first pass was not enough: **the live measurement in the description ran on a box with nothing mapped**, so the whole new code path was exercised only in its empty case. Configured a real map and found two things.

## With a real map, end to end

```
cfg_field_map   : firstname:FNAME,lastname:LNAME,gender:GENDER,dob:BIRTHDAY
cfg_field_map_n : 4
keys 26, 720 bytes
```

Real attribute ids resolved through the real `getMapFields()`, configuration restored afterwards.

## A pair that names nothing

An attribute mapped and later **deleted** leaves its id in the stored map, and the resolver returns a null attribute code for it:

```
cfg_field_map   : 'firstname:FNAME,:GHOST'
cfg_field_map_n : 2
warnings        : 6   ("Undefined array key 99999", "Trying to access array offset on null")
```

`:GHOST` is well formed enough to store and names nothing — the same shape as `__` on the surface field and `1_2_3` on the segments. A half-empty pair is now dropped, and the count reports what is carryable rather than what was stored.

**The six warnings are pre-existing** and come from `getMapFields()` indexing the attribute table without `isset()`. Not this method's to fix and not folded in — but they fire on every sync of an installation with a stale mapping, so they are worth their own look. Say if you want a ticket.

## And `array_filter()` without a callback

It drops an id of `"0"` along with the blanks. Mailchimp ids are not numeric today, and a filter that depends on that is one that becomes wrong quietly. Replaced with an explicit emptiness test.

## Tests

**263 tests, 582 assertions, green** — 3 more. Four mutations now, four killed:

| mutation | result |
|---|---|
| trim by characters instead of whole entries | 1 failure |
| report the trimmed count instead of the offered one | 1 failure |
| drop the half-empty-pair guard | 1 failure |
| put `array_filter()` back | 1 failure |

The description's "25 keys, 644 bytes" stays as what it was — one box with nothing mapped. The 26/720 above is the same box with four fields mapped, and neither is the contract, which is 28.

dipola (approved):
Approving, and thank you for asking rather than assuming — **the answer is no, and it fails in the opposite direction to the one you were worried about.**

You asked whether `cfg_field_map_n: 0` would count as a gate being off. It would. What would not work is the other value: **a non-zero count would also be treated as off**, so a live map would be cleared on every ping rather than kept. Your two boolean gates work; a numeric one does not, and it breaks the case that matters rather than the edge case.

**That is mine to fix, not yours — do not change the wire.** The count is the natural gate here: `cfg_field_map_n` already states how many, and an extra boolean would be a second copy of `n > 0` travelling on every ping to say something the number already says. Emitting it would make the wire worse to make my side easier, which is the wrong trade. I will take the four keys as this PR sends them.

Everything else reads correctly:

**Pairs rather than counts** — and the reason holds up: `firstname:FNAME` is what answers "first name is not syncing"; a count answers nothing. Both halves are identifiers rather than merchant content.

**Dropping whole entries against the serialised length** is the right bound, and for the precise reason you give — a list cut mid-pair arrives looking like a complete map that ends on `firstna`, which is the same class of well-formed-and-meaningless value as the bare delimiters earlier today. The offered count travelling beside the trimmed list is what makes a shortened map visibly shortened.

**Reading `XML_INTEREST` directly rather than through `Helper::getInterest()`** — worth having stated in the PR, since the resolving version is the obvious one to reach for and it would turn a configuration snapshot into an API call on the ping path.

**28 keys, 24 always plus 4 conditional, counted from `forStore()`** — that is the contract stated the way the last correction asked for, and the "25 keys, 644 bytes" is correctly labelled a sample of one box rather than the spec.

PR #2376, opened 2026-09-15
version 86 for magento 2.4

gonzaloebiz:
Version bump and changelog. No code changes. **The library floor is unchanged** — nothing here asks anything new of it, and `>=3.0.52` is already where it needs to be.

## ⚠ This release adds a database column

`mailchimp_sync_batches` gains `response_attempts`, so **`bin/magento setup:upgrade` is required** after updating. It is the first line of the changelog rather than a footnote: a merchant who updates without running it gets a site that refuses to serve.

## What it releases

**Two merchant-affecting fixes, not instrumentation:**

- **#2373 — the merge-field map was per process, not per store view.** `getMapFields()` remembered its result with no store id in the key, and the ecommerce cron walks every store view with one helper instance. Every view after the first synchronised with the **first view's map**: wrong merge fields written into the merchant's audience, nothing logged, no counter moved.
- **#2374 — the attribute option labels had the same defect one layer down**, found in review of the above. Narrower: only source-backed attributes, which on a stock install is `gender` plus any select attribute the merchant added.
- **#2371 — a batch result that cannot be read is no longer retried forever.** It was re-fetched and re-downloaded every cron run until Mailchimp expired the batch about a week later. Now five readings of a *finished* batch, then give up and re-send what it carried rather than deleting the sync rows. The two admin "get response" buttons also gained a distinct message for a batch that has not finished yet — previously indistinguishable from one Mailchimp had deleted.

**And the one that unlocks the rest:**

- **#2372 and #2375 — the configuration snapshot on the extension beacon.** The settings, the merge-field map and the interest groups. This is the only diagnostic the old support-log channel holds that nothing else replaces, and that channel cannot be switched off until this is landing. It carries request and response bodies — shopper emails and addresses — into the merchant's own database, and truncates them at 64 KB, which means it has been writing them broken.

No merchant-authored text travels on the snapshot: the two settings that are HTML the merchant writes are reported as whether they were customised, never as their content.

## Tests

**263 tests, 582 assertions, green.**

## Verification note

Two of the fixes above are for multi-store-view installations and this install has one store view, so their live reproduction is not in the record — the evidence for them is Magento's own source, measurements of the store ids involved, and unit tests with mutations. Stated here rather than left for the green number to imply.


dipola (approved):
Approving. `composer.json` and `setup_version` both moved, and the library floor correctly left at `>=3.0.52` — nothing in this release asks anything new of it.

**Putting the schema change in the first line of the changelog rather than in a footnote is the right call.** A merchant who updates without running `setup:upgrade` gets a site that refuses to serve, and that is not something to discover from a heading three sections down.

The column itself is the safe shape, which is worth saying since it is what makes the requirement merely procedural rather than risky:

```xml
<column xsi:type="smallint" name="response_attempts" padding="6" nullable="false" default="0"
```

`nullable="false"` with a default means every existing row in `mailchimp_sync_batches` gets `0` on the upgrade — no backfill, no NULL branch for the retry counter to handle, and a batch that predates the release starts its five-reading budget from the beginning rather than from an unknown state.

**The verification note is the part I would keep in the template for future releases.** Two of these fixes are for multi-store-view installations and this box has one store view, so their live reproduction is not in the record — and you said so rather than letting `263 tests, 582 assertions, green` imply otherwise. A green count that is allowed to stand for more than it covers is how the map memo survived for as long as it did: the tests that existed passed, and nobody had asked what they passed against.

**On what this unlocks** — the configuration snapshot is the only diagnostic the old support-log channel holds that nothing else replaces, so this is the release that makes switching that channel off possible at all. Worth knowing that the receiving end is already whitelisting all 28 keys, so the round trip becomes observable the moment an installation takes this.

